PR #45, opened 2021-02-08
Out of memory support

konraddysput:
This pull request allows backtrace-cocoa to handle out of memory exceptions that might happen after receiving a memory warning message generated by the environment when memory pressure is high. 

image

The diagram above shows what the oom algorithm does - with few modifications that we added to our codebase. On the application startup, an oom watcher creates a status file - with basic information about the application state. When the application state changes (application goes to background, out of memory warning occurred) the oom watcher updates the application state.

Basic information that we store in the application state file:
* application version,
* system version
* is debugger connected?
* application state - foreground vs background
* report.data - application state when we received a low memory warning
* report attributes - application attributes when we received a low memory warning

When the application status file is not available on application startup it means that user:
* updated backtrace-cocoa library,
* application crash occurred or user closed an application

Below you can find a summary of our algorithm:
1. On BacktraceOomWatcherInitialization validate if there is the application state file - if not, it means that user updated library or crash occurred or closed the application - so there is no reason to report out of memory,
2. If the application status exists, validate if the application was closed by the application update - if the application version is different, then the is no reason to report out of memory
3. If the application version is the same, validate the system version - if the system version is different - the system update occurred
4. If the application version and system version are the same validate if the debugger was connected on the last application session
5. If the debugger wasn't connected and report data is available - it means that we were able to capture the application state when low memory warning occurred and probably out of memory exception occured.
6. Before sending an error include attributes from the application status file to let the Backtrace user know if this was foreground out of exception or background out of exception.

Also - other our libraries report error.type attribute - in this diff you can find code responsible for assigning error.type attribute to backtrace-cocoa crashes.

OOM resources:
* https://engineering.fb.com/2015/08/24/ios/reducing-fooms-in-the-facebook-ios-app/
* https://developer.apple.com/documentation/xcode/improving_your_app_s_performance/reducing_your_app_s_memory_use/responding_to_low-memory_warnings
* https://github.com/trailbehind/TBOOMDetector


Sources/Features/Client/BacktraceReporter.swift
```diff
@@ -21,7 +22,9 @@ final class BacktraceReporter {
                              credentials: credentials,
                              repository: try PersistentRepository<BacktraceReport>(settings: dbSettings))
         self.repository = try PersistentRepository<BacktraceReport>(settings: dbSettings)
-        self.attributesProvider = AttributesProvider()
+        let localAttributeProvider = AttributesProvider()
```
rqbacktrace, line 25: What does `let` mean here?

Sources/Features/Client/BacktraceOomWatcher.swift
```diff
@@ -0,0 +1,148 @@
...
+        }
+        
+        // no low memory warning
+        if(backtraceOOMState["resource"] == nil) {
+            return
+        }
+        
+        // check if debugger was enabled
+        if(backtraceOOMState["debugger"] as! Bool == true) {
+            // detected system update
```
rqbacktrace, line 79: This should probably be deleted

Sources/Features/Client/BacktraceOomWatcher.swift
```diff
@@ -0,0 +1,148 @@
...
+            return
+        }
+        
+        var attributes = backtraceOOMState["resource.attributes"] as! Attributes
+        attributes["error.message"] = "Out of memory detected."
+        attributes["error.type"] = "Low Memory"
+        attributes["state"] = backtraceOOMState["state"]
+        
+        // ok - we detected oom and we should report it
+        guard let report = try? BacktraceReport(report: backtraceOOMState["resource"] as! Data, attributes: attributes,   attachmentPaths: []) else {
```
rqbacktrace, line 100: Very cool! Is there some way we could differentiate BOOM vs. FOOM here? I see we have hooks into background/foreground transition events in this pull request as well

Sources/Features/Attributes/AttributesProvider.swift
```diff
@@ -39,6 +39,10 @@ extension AttributesProvider: SignalContext {
         self.faultInfo.faultMessage = faultMessage
     }
     
+    func setErrorType(type: String?) {
```
rqbacktrace, line 42: Why do we need this field now?

rqbacktrace (approved):
This is so cool! Looks good to me!

Sources/Features/Client/BacktraceReporter.swift
```diff
@@ -96,3 +99,46 @@ extension BacktraceReporter {
...
+                                               name: UIApplication.willTerminateNotification,
+                                               object: nil)
+        
+        NotificationCenter.default.addObserver(self,
+                                               selector:#selector(handleLowMemoryWarning),
+                                               name: UIApplication.didReceiveMemoryWarningNotification,
+                                               object: nil)
+    }
+    
+    @objc private func applicationWillEnterForeground(_ notification: NSNotification) {
```
krml19, line 129:
```suggestion
    @objc private func applicationWillEnterForeground(_ notification: Notification) {
```

Sources/Features/Client/BacktraceReporter.swift
```diff
@@ -96,3 +99,46 @@ extension BacktraceReporter {
...
+        self.backtraceOomWatcher.applicationWillEnterForeground()
+    }
+    
+    
+    @objc private func didEnterBackgroundNotification(_ notification: NSNotification) {
```
krml19, line 134:
```suggestion
    @objc private func didEnterBackgroundNotification(_ notification: Notification) {
```

Sources/Features/Client/BacktraceReporter.swift
```diff
@@ -96,3 +99,46 @@ extension BacktraceReporter {
...
+        self.backtraceOomWatcher.didEnterBackgroundNotification()
+    }
+    
+    @objc func handleTermination() {
```
krml19, line 138: It probably can be private

Sources/Features/Client/BacktraceReporter.swift
```diff
@@ -96,3 +99,46 @@ extension BacktraceReporter {
...
+        self.backtraceOomWatcher.handleTermination()
+    }
+    @objc func handleLowMemoryWarning() {
```
krml19, line 141: It probably can be private

Sources/Public/BacktraceClient.swift
```diff
@@ -153,6 +153,10 @@ extension BacktraceClient: BacktraceReporting {
             return
         }
         
+        if(self.configuration.detectOOM) {
```
krml19, line 156:
```suggestion
        if self.configuration.detectOOM {
```

Sources/Features/Client/BacktraceOomWatcher.swift
```diff
@@ -0,0 +1,148 @@
+import Foundation
+
+final class BacktraceOomWatcher {
+    private var applicationState = [String :Any]()
```
krml19, line 4:
```suggestion
    private var applicationState = [String: Any]()
```

Sources/Features/Client/BacktraceOomWatcher.swift
```diff
@@ -0,0 +1,148 @@
...
+    
+    
+    private(set) var lowMemoryFilePrefix = "_lowMemory"
```
krml19, line 7: `private(set) var` means that the setter is private but the getter is internal. The rule of thumb: 
- if it never changes and can be defined in the compilation time: `static let` value
- if it never changes but it's a runtime value: `let` value
- if it can change: `private(set) var`

```suggestion
    let lowMemoryFilePrefix = "_lowMemory"
```

Sources/Features/Client/BacktraceOomWatcher.swift
```diff
@@ -0,0 +1,148 @@
...
+    private(set) static var oomFilePath: URL? = getStatusFilePath()
+    
+    private(set) var statusFilePath: URL
+    private(set) var crashReporter: CrashReporting
+    private(set) var attributesProvider:AttributesProvider
```
krml19, line 12:
```suggestion
    private(set) var attributesProvider: AttributesProvider
```

Sources/Features/Client/BacktraceOomWatcher.swift
```diff
@@ -0,0 +1,148 @@
...
+    init(repository:  PersistentRepository<BacktraceReport>, crashReporter: CrashReporting, attributes: AttributesProvider, backtraceApi: BacktraceApi) {
+        self.crashReporter = crashReporter
+        self.attributesProvider = attributes
+        self.backtraceApi = backtraceApi
+        self.repository = repository
+        
+        // set default state
+        self.applicationState["state"] = "foreground"
+        self.applicationState["debugger"] = DebuggerChecker.isAttached()
+        self.applicationState["appVersion"] = Bundle.main.object(forInfoDictionaryKey: kCFBundleVersionKey as String)
```
krml19, line 25: Is this casting to `String` necessary?

Sources/Features/Client/BacktraceOomWatcher.swift
```diff
@@ -0,0 +1,148 @@
...
+        self.applicationState["version"] = UIDevice.current.systemVersion
+        
+        // set status file url
+        if(BacktraceOomWatcher.oomFilePath == nil) {
```
krml19, line 29:
```suggestion
        if BacktraceOomWatcher.oomFilePath == nil {
```

krml19 (changes requested):
It's very good work but I'm going to temporarily block it because this code won't work on the macOS platform imho. Also, leaving a few comments for potential improvements.

Sources/Features/Attributes/AttributesProvider.swift
konraddysput: We're trying to followup conventions from other libraries to tell user about error type - in this case we set error type for crashes (error.type = crash) and oom - (low memory)

Sources/Features/Client/BacktraceOomWatcher.swift
```diff
@@ -0,0 +1,148 @@
...
+            // database path will point out sqlite database path
+            // oom status should exist in the same directory where database exists.
+            BacktraceOomWatcher.oomFilePath  = self.repository.url.deletingLastPathComponent().absoluteURL
```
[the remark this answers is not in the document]
konraddysput: done, thx

Sources/Features/Client/BacktraceOomWatcher.swift
```diff
@@ -0,0 +1,148 @@
...
+                .appendingPathComponent("BacktraceOOMState")
```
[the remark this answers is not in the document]
konraddysput: Removed local variable - now we have only one single static variable that we will update just in case if oomFilePath is nil

Sources/Features/Client/BacktraceOomWatcher.swift
konraddysput: No, it isn't. Good catch! Thanks :)

Sources/Features/Client/BacktraceOomWatcher.swift
konraddysput: done thanks!

Sources/Features/Client/BacktraceOomWatcher.swift
```diff
@@ -0,0 +1,148 @@
...
+            return
+        }
+        
+        // no low memory warning
+        if(backtraceOOMState["resource"] == nil) {
+            return
+        }
+        
+        // check if debugger was enabled
+        if(backtraceOOMState["debugger"] as! Bool == true) {
```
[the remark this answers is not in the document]
konraddysput: new application state model should fix this issue.

Sources/Features/Client/BacktraceOomWatcher.swift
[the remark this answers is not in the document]
konraddysput: Added new application state to the BacktraceOomWatcher - we won't use dictionary format anymore.

Sources/Public/BacktraceClient.swift
konraddysput: done, thx

Sources/Features/Client/BacktraceOomWatcher.swift
konraddysput: done, thx

Sources/Features/Client/BacktraceOomWatcher.swift
konraddysput: oh actually it is - kCFBundleVersionKey is CString. - we need to cast it to String I guess?

Sources/Features/Client/BacktraceOomWatcher.swift
```diff
@@ -0,0 +1,179 @@
...
+            return
+        }
+        // check system update
+        if appState.version != ProcessInfo.processInfo.operatingSystemVersionString {
+            // detected system update
+            return
+        }
+        
+        // check application update
+        if appState.appVersion != Bundle.main.object(forInfoDictionaryKey: kCFBundleVersionKey as String) as! String {
```
krml19, line 86: Please avoid force unwrapping

Sources/Features/Client/BacktraceOomWatcher.swift
```diff
@@ -0,0 +1,179 @@
...
+            return
+        }
+        
+        do {
+            _ = try backtraceApi.send(report)
+        } catch {
+            BacktraceLogger.error(error)
+            try? self.repository.save(report)
+        }
+        try! FileManager.default.removeItem(at: BacktraceOomWatcher.oomFilePath!)
```
krml19, line 104: Please avoid force unwrapping

Sources/Features/Client/BacktraceOomWatcher.swift
```diff
@@ -0,0 +1,179 @@
...
+            return
+        }
+        
+        // check application update
+        if appState.appVersion != Bundle.main.object(forInfoDictionaryKey: kCFBundleVersionKey as String) as! String {
+            // detected app update
+            return
+        }
+        
+        let attributes = try? JSONSerialization.jsonObject(with: appState.attributes!, options: []) as! Attributes
```
krml19, line 91: Please avoid force unwrapping

krml19, line 91: I suggest moving attributes serialization and report creation into the `do/catch` block and removing force unwrapping

Sources/Features/Client/BacktraceOomWatcher.swift
```diff
@@ -0,0 +1,179 @@
...
+            let previousAppState = try? decoder.decode(ApplicationInfo.self, from: data)
+            else { return nil }
+        return previousAppState
+        
+    }
+    
+    private func saveState() {
+        let encoder = PropertyListEncoder()
+        if let data = try? encoder.encode(self.state) {
+            let destPath = BacktraceOomWatcher.oomFilePath!.path
```
krml19, line 171: Please avoid force unwrapping

Sources/Features/Client/BacktraceReporter.swift
```diff
@@ -22,6 +23,11 @@ final class BacktraceReporter {
                              repository: try PersistentRepository<BacktraceReport>(settings: dbSettings))
         self.repository = try PersistentRepository<BacktraceReport>(settings: dbSettings)
         self.attributesProvider = AttributesProvider()
+        self.backtraceOomWatcher = BacktraceOomWatcher(
+            repository: self.repository,
+            crashReporter: self.reporter,
+            attributes: self.attributesProvider as! AttributesProvider,
```
krml19, line 29: Please avoid force unwrapping

Sources/Features/Client/BacktraceOomWatcher.swift
```diff
@@ -0,0 +1,179 @@
...
+            return
+        }
+        
+        // no low memory warning
+        if appState.resource == nil {
+            return
+        }
+        
+        // check if debugger was enabled
+        if appState.debugger == true {
```
krml19, line 75:
```suggestion
        if appState.debugger {
```

Sources/Features/Client/BacktraceOomWatcher.swift
```diff
@@ -0,0 +1,179 @@
...
+
+extension BacktraceOomWatcher {
+    func appChanedState(_ newState: ApplicationState) {
+        self.state.state = newState
+        saveState()
+    }
+    
+    func handleTermination() {
+        // application terminates correctly - for example: user decide to close app
+        try! FileManager.default.removeItem(atPath: BacktraceOomWatcher.oomFilePath!.path)
```
krml19, line 138: Please avoid force unwrapping

Sources/Features/Client/BacktraceOomWatcher.swift
```diff
@@ -0,0 +1,179 @@
...
+        
+        
+        self.state.attributes = try? JSONSerialization.data(withJSONObject: resource.attributes)
+        saveState()
+    }
+    
+    private func loadPreviousState() -> ApplicationInfo? {
+        let decoder = PropertyListDecoder()
+        
+        guard let data = try? Data.init(contentsOf: BacktraceOomWatcher.oomFilePath!),
```
krml19, line 161: Please avoid force unwrapping

krml19, line 161:
```suggestion
        guard let data = try? Data(contentsOf: BacktraceOomWatcher.oomFilePath!),
```

Sources/Features/Client/BacktraceOomWatcher.swift
```diff
@@ -0,0 +1,179 @@
...
+        return previousAppState
+        
+    }
+    
+    private func saveState() {
+        let encoder = PropertyListEncoder()
+        if let data = try? encoder.encode(self.state) {
+            let destPath = BacktraceOomWatcher.oomFilePath!.path
+            if FileManager.default.fileExists(atPath: destPath) {
+                try? data.write(to: BacktraceOomWatcher.oomFilePath!)
```
krml19, line 173: Please avoid force unwrapping

krml19 (commented):
Nice work! I strongly recommend removing force unwrapping because it is dangerous and might lead to crashes. It'd be also great to create throwing functions rather than functions that return an optional value. Having throwing functions the caller of the function is responsible for error handling. The caller can decide either to ignore the errors silently or handle them more gracefully.

Also, we can start looking into [a type of dispatch source for monitoring memory pressure events](https://developer.apple.com/documentation/dispatch/dispatch_source_type_memorypressure), so we can memory consumption on every platform.

Sources/Features/Client/BacktraceOomWatcher.swift
konraddysput: Done - replaced with the clean method

Sources/Features/Client/BacktraceOomWatcher.swift
konraddysput: avoided unwrapped - added one more if statement. However the one issue that I have with this implementation is that - we know this variable always exist - based on database implementation. We are always sure that oomFilePath will point to right place after runnning the class constructor.

Sources/Features/Client/BacktraceOomWatcher.swift
konraddysput: I have mixed feelings about it - if report is available but JSONSerializer throws an exception then in worst case we won't send useful data for user and fail because of optional values.

Sources/Features/Client/BacktraceReporter.swift
konraddysput: Swift compiler also warn me about var usage. I think we should stick with let - instead of var - but I'm also open for other suggestions.

Sources/Features/Client/BacktraceOomWatcher.swift
konraddysput: replaced with different definition

Sources/Features/Client/BacktraceOomWatcher.swift
konraddysput: done.

Sources/Features/Client/BacktraceOomWatcher.swift
konraddysput: done

Sources/Features/Client/BacktraceOomWatcher.swift
konraddysput: done.

Sources/Public/BacktraceClientConfiguration.swift
```diff
@@ -17,6 +17,8 @@ import Foundation
     /// Flag indicating if the Backtrace client should report reports when the debugger is attached. Default `false`.
     @objc public var allowsAttachingDebugger: Bool = false
     
+    /// Flag responsible for detecting and sending possible OOM cashes
+    @objc public var detectOOM: Bool = false;
```
krml19, line 21:
```suggestion
    @objc public var detectOOM: Bool = false
```

krml19, line 21: Please rename it to `detectsOOM` or `detectsOom`

Sources/Public/Internal/SignalContext.swift
```diff
@@ -4,4 +4,5 @@ protocol SignalContext: CustomStringConvertible {
     var allAttributes: Attributes { get }
     var attributes: Attributes { get set }
     func set(faultMessage: String?)
+    func setErrorType(type: String?)
```
krml19, line 7: I'd suggest renaming this method into `set(errorType: String?)` to make it more Swifty

Sources/Features/Client/BacktraceReporter.swift
```diff
@@ -21,7 +22,13 @@ final class BacktraceReporter {
                              credentials: credentials,
                              repository: try PersistentRepository<BacktraceReport>(settings: dbSettings))
         self.repository = try PersistentRepository<BacktraceReport>(settings: dbSettings)
-        self.attributesProvider = AttributesProvider()
+        let attributeProvider = AttributesProvider()
```
krml19, line 25: Please rename it to `attributesProvider`

Sources/Features/Client/BacktraceReporter.swift
```diff
@@ -96,3 +103,86 @@ extension BacktraceReporter {
...
+                                               object: nil)
+        #endif
+        
+        #if os(macOS)
+        let source = DispatchSource.makeMemoryPressureSource(eventMask: [.critical, .warning], queue: nil)
+        let dispatchQueue = DispatchQueue.init(label: "Backtrace")
+        dispatchQueue.async {
+            source.setEventHandler {
+                if source.mask == DispatchSource.MemoryPressureEvent.warning ||
+                    source.mask == DispatchSource.MemoryPressureEvent.critical {
```
krml19, line 157:
```suggestion
                if [.warning, .critical].contains(source.mask) {
```

Sources/Features/Client/BacktraceReporter.swift
```diff
@@ -96,3 +103,86 @@ extension BacktraceReporter {
...
+        
+        NotificationCenter.default.addObserver(self,
+                                               selector: #selector(handleLowMemoryWarning),
+                                               name: Application.didReceiveMemoryWarningNotification,
+                                               object: nil)
+        #endif
+        
+        #if os(macOS)
+        let source = DispatchSource.makeMemoryPressureSource(eventMask: [.critical, .warning], queue: nil)
+        let dispatchQueue = DispatchQueue.init(label: "Backtrace")
```
krml19, line 153:
```suggestion
        let dispatchQueue = DispatchQueue(label: "Backtrace OOM")
```

Sources/Features/Client/BacktraceReporter.swift
```diff
@@ -96,3 +103,86 @@ extension BacktraceReporter {
...
+        self.backtraceOomWatcher.start()
+        
+        NotificationCenter.default.addObserver(self,
+                                               selector: #selector(handleTermination),
+                                               name: Application.willTerminateNotification,
+                                               object: nil)
+        
+        NotificationCenter.default.addObserver(self,
+                                               selector: #selector(didBecomeActiveNotification),
+                                               name: Application.didBecomeActiveNotification, object: nil)
```
krml19, line 127: `object` can be in the next line to match the others

Sources/Features/Client/BacktraceReporter.swift
```diff
@@ -96,3 +103,86 @@ extension BacktraceReporter {
...
+        #endif
+        
+        #if os(macOS)
+        let source = DispatchSource.makeMemoryPressureSource(eventMask: [.critical, .warning], queue: nil)
+        let dispatchQueue = DispatchQueue.init(label: "Backtrace")
+        dispatchQueue.async {
+            source.setEventHandler {
+                if source.mask == DispatchSource.MemoryPressureEvent.warning ||
+                    source.mask == DispatchSource.MemoryPressureEvent.critical {
+                    self.handleLowMemoryWarning()
```
krml19, line 158: No action required. I understand that `Backtrace OOM` lives forever once initialized therefore it does not matter much. However, it's a good practice to use weak references within closures. 

```
      closure { [weak self] in 
           guard let self = self else { return } 
           self.handleLowMemoryWarning()
       }
```

Sources/Features/Client/BacktraceOomWatcher.swift
```diff
@@ -0,0 +1,220 @@
...
+            return nil
+        }
+        
+    }
+    
+    private func saveState() {
+        let encoder = PropertyListEncoder()
+        if let data = try? encoder.encode(self.state) {
+            if let destPath = BacktraceOomWatcher.oomFilePath {
+                if FileManager.default.fileExists(atPath: destPath.path) {
```
krml19, line 212: You can chain conditional expressions: 
```swift
        if let data = try? encoder.encode(self.state), let destPath = BacktraceOomWatcher.oomFilePath, FileManager.default.fileExists(atPath: destPath.path) { ...
     
```

Sources/Features/Client/BacktraceOomWatcher.swift
```diff
@@ -0,0 +1,220 @@
...
+        
+        guard let appState = self.loadPreviousState() else {
+            return
+        }
+        
+        if !shouldReportOom(appState: appState) {
+            return
+        }
+        
+
```
krml19, line 88: Too many new lines
```suggestion
```

Sources/Features/Client/BacktraceOomWatcher.swift
```diff
@@ -0,0 +1,220 @@
...
+        }
+        return cacheDir.appendingPathComponent("BacktraceOomState.plist")
+    }
+    
+    private func sendPendingOomReports() {
+        
+        // if oom state file doesn't exist it means that we deleted it to
+        // prevent sending false oom crashes
+        if let oomFilePath = BacktraceOomWatcher.oomFilePath {
+            if !FileManager.default.fileExists(atPath: oomFilePath.path) {
```
krml19, line 72: It can be a one-liner:

```swift
if let oomFilePath = BacktraceOomWatcher.oomFilePath, !FileManager.default.fileExists(atPath: oomFilePath.path) {
```

Sources/Features/Client/BacktraceOomWatcher.swift
```diff
@@ -0,0 +1,220 @@
...
+    }
+    
+    private func reportOom(appState: ApplicationInfo) {
+        guard let reportData = appState.resource else {
+            return
+        }
+        
+        var reportAttributes: Attributes = [:]
+        if let stateAttributes = appState.attributes {
+            if let attributes = try? JSONSerialization.jsonObject(with: stateAttributes, options: []) as? Attributes {
```
krml19, line 126: It can be a one-liner: 
```swift
if let stateAttributes = appState.attributes, let attributes = try? JSONSerialization.jsonObject(with: stateAttributes, options: []) as? Attributes {
```

Sources/Features/Client/BacktraceOomWatcher.swift
```diff
@@ -0,0 +1,220 @@
...
+        let decoder = PropertyListDecoder()
+        
+        if let destPath = BacktraceOomWatcher.oomFilePath {
+            guard let data = try? Data.init(contentsOf: destPath),
+                let previousAppState = try? decoder.decode(ApplicationInfo.self, from: data)
+                else { return nil }
+            return previousAppState
+        } else {
+            return nil
+        }
```
krml19, line 204: It can be check with one expression: 

```swift
        guard let destPath = BacktraceOomWatcher.oomFilePath, let data = try? Data(contentsOf: destPath), let previousAppState = try? decoder.decode(ApplicationInfo.self, from: data) else { return nil }
            return previousAppState
        }
```

krml19 (commented):
Great work!

krml19 (approved):
LGTM 👍

Sources/Features/Client/BacktraceOomWatcher.swift
krml19: ok, fine

PR #59, opened 2021-04-07
Attachment support for crash reports

rqbacktrace:
This adds attachment support functionality for crash reports on `backtrace-cocoa`

I added this functionality to the demo app as well and manually tested live and crash reports and see the correct file attachment in both cases. I also manually tested live and crash reports without file attachments and they both still submitted correctly with no file attachments.

This does not fix a currently known bug where live-report specific attachments are not saved if that live-report is submitted on a retry. We can potentially use the functionality added here to fix that bug.

@KrzaQ I tagged you because I used your approach adding file attachment support to Crashpad, you might be better able to spot pitfalls in this approach

### DIFF
--- a/Examples/Example-iOS/AppDelegate.swift
+++ b/Examples/Example-iOS/AppDelegate.swift
@@ -18,6 +18,7 @@ final class AppDelegate: UIResponder, UIApplicationDelegate {
                      didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
         let backtraceCredentials = BacktraceCredentials(endpoint: URL(string: Keys.backtraceUrl as String)!,
                                                         token: Keys.backtraceToken as String)
+
         let backtraceDatabaseSettings = BacktraceDatabaseSettings()
         backtraceDatabaseSettings.maxRecordCount = 1000
         backtraceDatabaseSettings.maxDatabaseSize = 10
@@ -69,11 +70,7 @@ final class AppDelegate: UIResponder, UIApplicationDelegate {
         let formatter = DateFormatter()
         formatter.timeStyle = .medium
         let myData = formatter.string(from: Date())
-        do {
-            try myData.write(to: fileUrl, atomically: true, encoding: .utf8)
-        } catch {
-            print("Error: \(error)")
-        }
+        try myData.write(to: fileUrl, atomically: true, encoding: .utf8)
         return fileUrl
     }
 }
--- a/Examples/Example-iOS/AppDelegate.swift
+++ b/Examples/Example-iOS/AppDelegate.swift
@@ -14,20 +14,6 @@ final class AppDelegate: UIResponder, UIApplicationDelegate {
     
     var window: UIWindow?
     
-    public func createAndWriteFile(_ fileUrl: URL) {
-        // prints the file path
-        print("File path \(fileUrl.path)")
-        //data to write in file.
-        let formatter = DateFormatter()
-        formatter.timeStyle = .medium
-        let stringData = formatter.string(from: Date())
-        do {
-            try stringData.write(to: fileUrl, atomically: true, encoding: String.Encoding.utf8)
-        } catch let error as NSError {
-            print (error)
-        }
-    }
-    
     func application(_ application: UIApplication,
                      didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
         let backtraceCredentials = BacktraceCredentials(endpoint: URL(string: Keys.backtraceUrl as String)!,
@@ -47,22 +33,12 @@ final class AppDelegate: UIResponder, UIApplicationDelegate {
         BacktraceClient.shared = try? BacktraceClient(configuration: backtraceConfiguration)
         BacktraceClient.shared?.delegate = self
         BacktraceClient.shared?.attributes = ["foo": "bar", "testing": true]
+        
         let fileName = "sample.txt"
-        let dirName = "directory"
-        let libraryDirectoryUrl = try! FileManager.default.url(
-           for: .libraryDirectory, in: .userDomainMask, appropriateFor: nil, create: true
-           )
-        let directoryUrl = libraryDirectoryUrl.appendingPathComponent(dirName)
-        try? FileManager().createDirectory(
-                    at: directoryUrl,
-                    withIntermediateDirectories: false,
-                    attributes: nil
-                )
-        let fileUrl = directoryUrl.appendingPathComponent(fileName)
+        let fileUrl = try? createAndWriteFile(fileName)
         var crashAttachments = Attachments()
-        crashAttachments[fileName] = fileUrl as URL
+        crashAttachments[fileName] = fileUrl
         BacktraceClient.shared?.attachments = crashAttachments
-        createAndWriteFile(fileUrl)
 
         BacktraceClient.shared?.loggingDestinations = [BacktraceBaseDestination(level: .debug)]
         do {
@@ -75,6 +51,30 @@ final class AppDelegate: UIResponder, UIApplicationDelegate {
         }
         
         return true
+    }
+    
+    func createAndWriteFile(_ fileName: String) throws -> URL {
+        let dirName = "directory"
+        guard let libraryDirectoryUrl = try? FileManager.default.url(
+            for: .libraryDirectory, in: .userDomainMask, appropriateFor: nil, create: true) else {
+            throw CustomError.runtimeError
+        }
+        let directoryUrl = libraryDirectoryUrl.appendingPathComponent(dirName)
+        try? FileManager().createDirectory(
+                    at: directoryUrl,
+                    withIntermediateDirectories: false,
+                    attributes: nil
+                )
+        let fileUrl = directoryUrl.appendingPathComponent(fileName)
+        let formatter = DateFormatter()
+        formatter.timeStyle = .medium
+        let myData = formatter.string(from: Date())
+        do {
+            try myData.write(to: fileUrl, atomically: true, encoding: .utf8)
+        } catch {
+            print("Error: \(error)")
+        }
+        return fileUrl
     }
 }
 
--- a/Examples/Example-iOS/AppDelegate.swift
+++ b/Examples/Example-iOS/AppDelegate.swift
@@ -14,6 +14,20 @@ final class AppDelegate: UIResponder, UIApplicationDelegate {
     
     var window: UIWindow?
     
+    public func createAndWriteFile(_ fileUrl: URL) {
+        // prints the file path
+        print("File path \(fileUrl.path)")
+        //data to write in file.
+        let formatter = DateFormatter()
+        formatter.timeStyle = .medium
+        let stringData = formatter.string(from: Date())
+        do {
+            try stringData.write(to: fileUrl, atomically: true, encoding: String.Encoding.utf8)
+        } catch let error as NSError {
+            print (error)
+        }
+    }
+    
     func application(_ application: UIApplication,
                      didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
         let backtraceCredentials = BacktraceCredentials(endpoint: URL(string: Keys.backtraceUrl as String)!,
@@ -24,7 +38,7 @@ final class AppDelegate: UIResponder, UIApplicationDelegate {
         backtraceDatabaseSettings.retryInterval = 5
         backtraceDatabaseSettings.retryLimit = 3
         backtraceDatabaseSettings.retryBehaviour = RetryBehaviour.interval
-        backtraceDatabaseSettings.retryOrder = RetryOder.queue
+        backtraceDatabaseSettings.retryOrder = RetryOrder.queue
         let backtraceConfiguration = BacktraceClientConfiguration(credentials: backtraceCredentials,
                                                                   dbSettings: backtraceDatabaseSettings,
                                                                   reportsPerMin: 10,
@@ -33,7 +47,23 @@ final class AppDelegate: UIResponder, UIApplicationDelegate {
         BacktraceClient.shared = try? BacktraceClient(configuration: backtraceConfiguration)
         BacktraceClient.shared?.delegate = self
         BacktraceClient.shared?.attributes = ["foo": "bar", "testing": true]
-        
+        let fileName = "sample.txt"
+        let dirName = "directory"
+        let libraryDirectoryUrl = try! FileManager.default.url(
+           for: .libraryDirectory, in: .userDomainMask, appropriateFor: nil, create: true
+           )
+        let directoryUrl = libraryDirectoryUrl.appendingPathComponent(dirName)
+        try? FileManager().createDirectory(
+                    at: directoryUrl,
+                    withIntermediateDirectories: false,
+                    attributes: nil
+                )
+        let fileUrl = directoryUrl.appendingPathComponent(fileName)
+        var crashAttachments = Attachments()
+        crashAttachments[fileName] = fileUrl as URL
+        BacktraceClient.shared?.attachments = crashAttachments
+        createAndWriteFile(fileUrl)
+
         BacktraceClient.shared?.loggingDestinations = [BacktraceBaseDestination(level: .debug)]
         do {
             try throwingFunc()
--- a/Sources/Features/Attributes/AttributesProvider.swift
+++ b/Sources/Features/Attributes/AttributesProvider.swift
@@ -44,12 +44,8 @@ extension AttributesProvider: SignalContext {
         self.attributes["error.type"] = errorType
     }
     
-    var attachmentPathsArray: [ String ] {
-        var attachmentPathsArray = [ String ]()
-        for attachment in attachments {
-            attachmentPathsArray.append(attachment.value.path)
-        }
-        return attachmentPathsArray
+    var attachmentPathsArray: [String] {
+        return attachments.map(\.value.path)
     }
     
     var allAttributes: Attributes {
--- a/Sources/Features/Attributes/AttributesProvider.swift
+++ b/Sources/Features/Attributes/AttributesProvider.swift
@@ -44,7 +44,7 @@ extension AttributesProvider: SignalContext {
         self.attributes["error.type"] = errorType
     }
     
-    var attachmentPathsArray: [String] {
+    var attachmentPaths: [String] {
         return attachments.map(\.value.path)
     }
     
--- a/Sources/Features/Attributes/AttributesProvider.swift
+++ b/Sources/Features/Attributes/AttributesProvider.swift
@@ -14,6 +14,7 @@ final class AttributesProvider {
     
     // attributes can be modified on runtime
     var attributes: Attributes = [:]
+    var attachments: Attachments = [:]
     private let attributesSources: [AttributesSource]
     private let faultInfo: FaultInfo
     
@@ -41,6 +42,14 @@ extension AttributesProvider: SignalContext {
     
     func set(errorType: String?) {
         self.attributes["error.type"] = errorType
+    }
+    
+    var attachmentPathsArray: [ String ] {
+        var attachmentPathsArray = [ String ]()
+        for attachment in attachments {
+            attachmentPathsArray.append(attachment.value.path)
+        }
+        return attachmentPathsArray
     }
     
     var allAttributes: Attributes {
--- a/Sources/Features/Client/BacktraceReporter.swift
+++ b/Sources/Features/Client/BacktraceReporter.swift
@@ -78,6 +78,14 @@ extension BacktraceReporter: BacktraceClientCustomizing {
             attributesProvider.attributes = newValue
         }
     }
+    
+    var attachments: Attachments {
+        get {
+            return attributesProvider.attachments
+        } set {
+            attributesProvider.attachments = newValue
+        }
+    }
 }
 
 extension BacktraceReporter {
@@ -96,7 +104,7 @@ extension BacktraceReporter {
         attributesProvider.set(faultMessage: faultMessage)
         let resource = try reporter.generateLiveReport(exception: exception,
                                                        attributes: attributesProvider.allAttributes,
-                                                       attachmentPaths: attachmentPaths)
+                                                       attachmentPaths: attachmentPaths + attributesProvider.attachmentPathsArray)
         return send(resource: resource)
     }
     
@@ -106,7 +114,7 @@ extension BacktraceReporter {
         attributesProvider.set(errorType: "Exception")
         let resource = try reporter.generateLiveReport(exception: exception,
                                                        attributes: attributesProvider.allAttributes,
-                                                       attachmentPaths: attachmentPaths)
+                                                       attachmentPaths: attachmentPaths + attributesProvider.attachmentPathsArray)
         return resource
     }
 }
--- a/Sources/Features/Client/BacktraceReporter.swift
+++ b/Sources/Features/Client/BacktraceReporter.swift
@@ -104,7 +104,7 @@ extension BacktraceReporter {
         attributesProvider.set(faultMessage: faultMessage)
         let resource = try reporter.generateLiveReport(exception: exception,
                                                        attributes: attributesProvider.allAttributes,
-                                                       attachmentPaths: attachmentPaths + attributesProvider.attachmentPathsArray)
+                                                       attachmentPaths: attachmentPaths + attributesProvider.attachmentPaths)
         return send(resource: resource)
     }
     
@@ -114,7 +114,7 @@ extension BacktraceReporter {
         attributesProvider.set(errorType: "Exception")
         let resource = try reporter.generateLiveReport(exception: exception,
                                                        attributes: attributesProvider.allAttributes,
-                                                       attachmentPaths: attachmentPaths + attributesProvider.attachmentPathsArray)
+                                                       attachmentPaths: attachmentPaths + attributesProvider.attachmentPaths)
         return resource
     }
 }
--- a/Sources/Features/Client/Model/AttachmentsStorage.swift
+++ b/Sources/Features/Client/Model/AttachmentsStorage.swift
@@ -1,9 +1,5 @@
 import Foundation
 
-/// Type-alias of storing file attachments on disk (as a bookmark)
-/// Expected format: Filename, File URL bookmark
-private typealias Bookmarks = [String: Data]
-
 enum AttachmentsStorageError: Error {
     case invalidDictionary
     case invalidBookmark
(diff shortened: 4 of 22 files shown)


Sources/Public/Internal/SignalContext.swift
```diff
@@ -2,7 +2,10 @@ import Foundation
 
 protocol SignalContext: CustomStringConvertible {
     var allAttributes: Attributes { get }
+    // We use attachment paths as strings in our report send functions
+    var attachmentPathsArray: [ String ] { get }
     var attributes: Attributes { get set }
+    var attachments: Attachments { get set }
```
konraddysput, line 8: ` our report send functions` <- report send function.. responsible for sending crashes? 
Can we also add description to `attachments` variable to avoid confussion later?

Tests/AttachmentStorageTests.swift
```diff
@@ -0,0 +1,60 @@
...
+import Quick
+@testable import Backtrace
+
+final class AttachmentStorageTests: QuickSpec {
+    
+    override func spec() {
+        describe("AttachmentStorage") {
+            it("can save attachments as a plist") {
+                var crashAttachments = Attachments()
+                guard let fileUrl = try? self.createAFile() else {
```
konraddysput, line 13: Do we create physical file on hard drive?

konraddysput (changes requested):
Hey - can we rethink attachmentStorage logic?

Examples/Example-iOS/AppDelegate.swift
```diff
@@ -33,7 +47,23 @@ final class AppDelegate: UIResponder, UIApplicationDelegate {
         BacktraceClient.shared = try? BacktraceClient(configuration: backtraceConfiguration)
         BacktraceClient.shared?.delegate = self
         BacktraceClient.shared?.attributes = ["foo": "bar", "testing": true]
-        
+        let fileName = "sample.txt"
```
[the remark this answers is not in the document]
rqbacktrace: You are right, I will make this a separate function to make it more clear for users how to use our `backtrace-cocoa` library

Sources/Features/Client/Model/AttachmentsStorage.swift
```diff
@@ -0,0 +1,131 @@
...
+        BacktraceLogger.debug("Removed attachments plist at path: \(config.fileUrl)")
+    }
+    
+    private static func convertAttachmentUrlsToBookmarks(_ attachments: Attachments) throws -> Bookmarks {
+        var attachmentsBookmarksDict = Bookmarks()
+        for attachment in attachments {
+            do { let bookmark = try attachment.value.bookmarkData(options: URL.BookmarkCreationOptions.minimalBookmark)
+                attachmentsBookmarksDict[attachment.key] = bookmark as NSData
+            } catch {
+                print("Caught error: \(error)")
```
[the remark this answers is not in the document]
rqbacktrace: Ah thank you for catching this, I will clean this up!

Sources/Public/BacktraceCrashReporter.swift
```diff
@@ -52,19 +56,49 @@ extension BacktraceCrashReporter: CrashReporting {
...
     }
     
+    // This function retrieves, constructs, and sends the pending crash report
     func pendingCrashReport() throws -> BacktraceReport {
         let reportData = try reporter.loadPendingCrashReportDataAndReturnError()
         let attributes = (try? AttributesStorage.retrieve(fileName: BacktraceCrashReporter.crashName)) ?? [:]
-        // NOTE: - no attachments in crash reports
-        return try BacktraceReport(report: reportData, attributes: attributes, attachmentPaths: [])
+        let attachmentPaths = copiedFileAttachments.map {$0.path}
+        return try BacktraceReport(report: reportData, attributes: attributes, attachmentPaths: attachmentPaths)
```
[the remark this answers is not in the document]
rqbacktrace: I don't think so. The crash report will contain a plist with its own attachment paths from the previous session. If we add attachments user declared in the current user session, we will potentially send double attachments, one from previous session and one from current user session. It could be a cool feature but maybe it can cause confusion

Sources/Public/BacktraceCrashReporter.swift
```diff
@@ -52,19 +56,49 @@ extension BacktraceCrashReporter: CrashReporting {
...
     }
     
+    // This function is called to copy stored file attachments
+    // from pending crashes so that they are not overwritten by the
+    // new app session
+    func copyFileAttachmentsFromPendingCrashes() -> [URL] {
+        guard let cacheDirectoryUrl = FileManager.default.urls(for: .cachesDirectory, in: .userDomainMask).first else {
```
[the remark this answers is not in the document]
rqbacktrace: That's a good idea, I will try it!

Sources/Public/BacktraceCrashReporter.swift
```diff
@@ -52,19 +56,49 @@ extension BacktraceCrashReporter: CrashReporting {
...
+            let fileManager = FileManager()
+            let copiedAttachmentPath = cacheDirectoryUrl.appendingPathComponent(attachment.key)
+            do {
+                if fileManager.fileExists(atPath: copiedAttachmentPath.path) {
+                    try fileManager.removeItem(atPath: copiedAttachmentPath.path)
+                }
+                try fileManager.copyItem(at: attachment.value, to: copiedAttachmentPath)
+                copiedFileAttachments.append(copiedAttachmentPath)
+            } catch {
+                print("Caught error: \(error)")
```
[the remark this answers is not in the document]
rqbacktrace: Ah thank you, will remove!

Sources/Public/BacktraceCrashReporter.swift
```diff
@@ -52,19 +56,49 @@ extension BacktraceCrashReporter: CrashReporting {
...
+        let attachments = (try? AttachmentsStorage.retrieve(fileName: BacktraceCrashReporter.crashName)) ?? [:]
+        var copiedFileAttachments = [URL]()
+        for attachment in attachments {
+            let fileManager = FileManager()
+            let copiedAttachmentPath = cacheDirectoryUrl.appendingPathComponent(attachment.key)
+            do {
+                if fileManager.fileExists(atPath: copiedAttachmentPath.path) {
+                    try fileManager.removeItem(atPath: copiedAttachmentPath.path)
+                }
+                try fileManager.copyItem(at: attachment.value, to: copiedAttachmentPath)
```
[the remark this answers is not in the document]
rqbacktrace: That's an awesome point, thank you Konrad!

Sources/Public/Internal/SignalContext.swift
rqbacktrace: I will clean this up, thank you!

Tests/AttachmentStorageTests.swift
rqbacktrace: Yes, we are creating a physical file on the hard drive here.

Sources/Features/Client/Model/AttachmentsStorage.swift
```diff
@@ -0,0 +1,131 @@
...
+/// Expected format: Filename, File URL bookmark
+public typealias Bookmarks = [String: NSData]
+
+final class AttachmentsStorage {
+    struct Config {
+        let cacheUrl: URL
+        let directoryUrl: URL
+        let fileUrl: URL
+        
+        init(fileName: String) throws {
```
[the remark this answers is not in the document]
rqbacktrace: You're right...let me think how I can better unify this with our existing `AttributesStorage.swift`

Sources/Public/BacktraceCrashReporter.swift
```diff
@@ -52,19 +56,49 @@ extension BacktraceCrashReporter: CrashReporting {
...
+    // new app session
+    func copyFileAttachmentsFromPendingCrashes() -> [URL] {
+        guard let cacheDirectoryUrl = FileManager.default.urls(for: .cachesDirectory, in: .userDomainMask).first else {
+            BacktraceLogger.error("Could not get cache directory URL")
+            return [URL]()
+        }
+        let attachments = (try? AttachmentsStorage.retrieve(fileName: BacktraceCrashReporter.crashName)) ?? [:]
+        var copiedFileAttachments = [URL]()
+        for attachment in attachments {
+            let fileManager = FileManager()
```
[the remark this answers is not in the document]
rqbacktrace: I don't think it will work, cacheDirectoryUrl calls a static function of FileManager (`FileManager.default.urls`). Otherwise for our file operations below we need an actual FileManager instance.

Sources/Features/Client/Model/AttachmentsStorage.swift
```diff
@@ -0,0 +1,131 @@
+import Foundation
+
+/// Type-alias of storing file attachments on disk (as a bookmark)
+/// Expected format: Filename, File URL bookmark
+public typealias Bookmarks = [String: NSData]
+
+final class AttachmentsStorage {
```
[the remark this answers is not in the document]
rqbacktrace: It won't be easy to unify these since the function parameters are a bit different. Even the Configs are different (`AttachmentsStorage.swift` needs a different plist path than `AttributesStorage.swift`). Even so, I think it can be done, but it will involve significant refactoring. It's probably a good idea for us to do this, especially if we want to add other kinds of `Storage` classes for other kinds of report metadata.

Sources/Features/Client/Model/AttachmentsStorage.swift
```diff
@@ -0,0 +1,129 @@
...
+        } else {
+            guard let dictionaryFromFile = NSDictionary(contentsOf: config.fileUrl) else {
+                throw FileError.invalidPropertyList
+            }
+            dictionary = dictionaryFromFile
+        }
+        
+        guard let bookmarks = dictionary as? Bookmarks else {
+            BacktraceLogger.debug("Could not convert stored dictionary to Bookmarks type")
+            throw FileError.unsupportedScheme
```
krml19, line 65: The thrown error does not match the failure reason.

Sources/Features/Client/Model/AttachmentsStorage.swift
```diff
@@ -0,0 +1,129 @@
...
+        }
+        guard let attachments = try? extractAttachmentUrls(bookmarks) else {
+            BacktraceLogger.debug("Could not extract attachment URLs from stored attachments Bookmarks")
+            throw FileError.invalidPropertyList
```
krml19, line 69: The thrown error does not match the failure reason.

Sources/Features/Client/Model/AttachmentsStorage.swift
```diff
@@ -0,0 +1,129 @@
...
+    
+    static func retrieveAsArray(fileName: String) throws -> [String] {
+        guard let attachments = try? retrieve(fileName: fileName) else {
+            return []
+        }
+        var attachmentsArray = [ String ]()
+        for attachment in attachments {
+            attachmentsArray.append(attachment.value.path)
+        }
+        return attachmentsArray
```
krml19, line 84: Saw it in other places as well. You might want to think of adding a method or a computed property to get this value 

```suggestion
let attachmentsArray = attachments.map(\.value.path)
```

Sources/Features/Client/Model/AttachmentsStorage.swift
```diff
@@ -0,0 +1,129 @@
...
+        }
+        // remove file
+        try FileManager.default.removeItem(at: config.fileUrl)
+        BacktraceLogger.debug("Removed attachments plist at path: \(config.fileUrl)")
+    }
+    
+    private static func convertAttachmentUrlsToBookmarks(_ attachments: Attachments) throws -> Bookmarks {
+        var attachmentsBookmarksDict = Bookmarks()
+        for attachment in attachments {
+            do { let bookmark = try attachment.value.bookmarkData(options: URL.BookmarkCreationOptions.minimalBookmark)
```
krml19, line 101: Expression should start from the new line

Sources/Features/Client/Model/AttachmentsStorage.swift
```diff
@@ -0,0 +1,129 @@
...
+            var stale = ObjCBool(false)
+            guard let fileUrl = try? NSURL.init(resolvingBookmarkData: bookmark.value as Data,
+                                                options: NSURL.BookmarkResolutionOptions(),
+                                                relativeTo: nil,
+                                                bookmarkDataIsStale: &stale) else {
+                BacktraceLogger.error("Could not resolve file URL from bookmark")
+                continue
+            }
+            if stale.boolValue {
+                BacktraceLogger.error("Bookmark data is stale. This should not happen")
```
krml19, line 123: The error seems to be silently ignored. Is it intentional?

Sources/Features/Client/Model/AttachmentsStorage.swift
```diff
@@ -0,0 +1,129 @@
...
+            }
+        }
+        return attachmentsBookmarksDict
+    }
+    
+    private static func extractAttachmentUrls(_ bookmarks: Bookmarks) throws -> Attachments {
+        var attachments = Attachments()
+        for bookmark in bookmarks {
+            var stale = ObjCBool(false)
+            guard let fileUrl = try? NSURL.init(resolvingBookmarkData: bookmark.value as Data,
```
krml19, line 115:
```suggestion
            guard let fileUrl = try? NSURL(resolvingBookmarkData: bookmark.value as Data,
```
Can we replace `NSURL` with `URL`?

krml19 (changes requested):
Please adjust the style

Tests/AttachmentStorageTests.swift
konraddysput: Are you sure we should create file in unit tests project? Unit tests shouldn't create files on the hard drive. @krml19 what do you think?

Sources/Public/BacktraceCrashReporter.swift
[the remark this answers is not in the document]
konraddysput: `FileManager.default` is your Filemanager instance.

Sources/Features/Attributes/AttributesProvider.swift
```diff
@@ -43,6 +44,14 @@ extension AttributesProvider: SignalContext {
         self.attributes["error.type"] = errorType
     }
     
+    var attachmentPathsArray: [ String ] {
+        var attachmentPathsArray = [ String ]()
+        for attachment in attachments {
+            attachmentPathsArray.append(attachment.value.path)
+        }
+        return attachmentPathsArray
+    }
```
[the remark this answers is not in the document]
rqbacktrace: Will do, thank you!

Sources/Features/Client/Model/AttachmentsStorage.swift
```diff
@@ -0,0 +1,129 @@
+import Foundation
+
+/// Type-alias of storing file attachments on disk (as a bookmark)
+/// Expected format: Filename, File URL bookmark
+public typealias Bookmarks = [String: NSData]
```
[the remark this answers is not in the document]
rqbacktrace: Will make these changes, thank you!

Sources/Public/Internal/SignalContext.swift
```diff
@@ -3,6 +3,10 @@ import Foundation
 protocol SignalContext: CustomStringConvertible {
     var allAttributes: Attributes { get }
     var attributes: Attributes { get set }
+    // File attachments are stored to disk as URLs
+    var attachments: Attachments { get set }
+    // File attachments are used in BacktraceReports as string paths
+    var attachmentPathsArray: [ String ] { get }
```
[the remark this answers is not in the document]
rqbacktrace: Will do, thank you!

Sources/Public/Internal/SignalContext.swift
```diff
@@ -3,6 +3,10 @@ import Foundation
 protocol SignalContext: CustomStringConvertible {
     var allAttributes: Attributes { get }
     var attributes: Attributes { get set }
+    // File attachments are stored to disk as URLs
+    var attachments: Attachments { get set }
+    // File attachments are used in BacktraceReports as string paths
```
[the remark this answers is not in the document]
rqbacktrace: Will do, thank you!

Sources/Features/Client/Model/AttachmentsStorage.swift
```diff
@@ -0,0 +1,129 @@
...
+
+        BacktraceLogger.debug("Retrieved attachments from path: \(config.fileUrl)")
+        return attachments
+    }
+    
+    static func retrieveAsArray(fileName: String) throws -> [String] {
+        guard let attachments = try? retrieve(fileName: fileName) else {
+            return []
+        }
+        var attachmentsArray = [ String ]()
```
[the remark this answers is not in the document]
rqbacktrace: Will do, thank you!

Sources/Features/Client/Model/AttachmentsStorage.swift
rqbacktrace: I will create a new error type for these errors, thank you!

Sources/Features/Client/Model/AttachmentsStorage.swift
rqbacktrace: I will create a new error type for these errors, thank you!

Sources/Features/Client/Model/AttachmentsStorage.swift
rqbacktrace: I think I can remove this function, it was only used in tests anyway. Thank you for showing me how to use `map` function, this is very helpful for writing clean Swift code.

Sources/Features/Client/Model/AttachmentsStorage.swift
rqbacktrace: Will fix, thank you!

Sources/Features/Client/Model/AttachmentsStorage.swift
rqbacktrace: Yes we can, thank you!

Sources/Features/Client/Model/AttachmentsStorage.swift
rqbacktrace: We should handle this error since we're not supposed to use stale URLs. Thank you for pointing this out!

Sources/Features/Client/Model/AttachmentsStorage.swift
```diff
@@ -0,0 +1,129 @@
+import Foundation
+
+/// Type-alias of storing file attachments on disk (as a bookmark)
+/// Expected format: Filename, File URL bookmark
+public typealias Bookmarks = [String: NSData]
+
+final class AttachmentsStorage {
```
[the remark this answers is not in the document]
rqbacktrace: Will do, thank you! I will make this change to `AttributesStorage.swift` as well!

Sources/Public/BacktraceCrashReporter.swift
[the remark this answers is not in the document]
rqbacktrace: We don't use `FileManager.default` here anymore. I agree with your feedback that it's better to get this path directly from `AttachmentsStorage`.

Tests/AttachmentStorageTests.swift
krml19: Yes, I agree. It'd be definitely better to use a mock

Sources/Public/Internal/SignalContext.swift
```diff
@@ -3,6 +3,10 @@ import Foundation
 protocol SignalContext: CustomStringConvertible {
     var allAttributes: Attributes { get }
     var attributes: Attributes { get set }
+    // File attachments are stored to disk as URLs
+    var attachments: Attachments { get set }
+    // File attachments are used in `BacktraceReport` as string paths
+    var attachmentPathsArray: [String] { get }
```
krml19, line 9: What's the reason for storing `attachments` and `attachmentPathsArray`? Also, I suggest renaming `attachementPathsArray` to `attachmentPaths`.

Sources/Public/BacktraceCrashReporter.swift
```diff
@@ -52,19 +56,62 @@ extension BacktraceCrashReporter: CrashReporting {
...
     
     func purgePendingCrashReport() throws {
         try AttributesStorage.remove(fileName: BacktraceCrashReporter.crashName)
+        try AttachmentsStorage.remove(fileName: BacktraceCrashReporter.crashName)
+        try deleteCopiedFileAttachments()
         try reporter.purgePendingCrashReportAndReturnError()
     }
+    
+    func deleteCopiedFileAttachments() throws {
+        let fileManager = FileManager()
```
krml19, line 110:
```suggestion
        let fileManager = FileManager.default
```

Sources/Features/Client/Model/AttachmentsStorage.swift
```diff
@@ -0,0 +1,125 @@
...
+    }
+    
+    private static func extractAttachmentUrls(_ bookmarks: Bookmarks) throws -> Attachments {
+        var attachments = Attachments()
+        for bookmark in bookmarks {
+            var stale = Bool(false)
+            guard let fileUrl = try? URL(resolvingBookmarkData: bookmark.value,
+                                                options: URL.BookmarkResolutionOptions(),
+                                                relativeTo: nil,
+                                                bookmarkDataIsStale: &stale) else {
```
krml19, line 113: Please align all parameters

Sources/Features/Client/BacktraceReporter.swift
```diff
@@ -96,7 +104,7 @@ extension BacktraceReporter {
         attributesProvider.set(faultMessage: faultMessage)
         let resource = try reporter.generateLiveReport(exception: exception,
                                                        attributes: attributesProvider.allAttributes,
-                                                       attachmentPaths: attachmentPaths)
+                                                       attachmentPaths: attachmentPaths + attributesProvider.attachmentPathsArray)
```
rqbacktrace, line 107: @krml19 this is why we need `attachmentPathsArray` parameter. But I will change it to `attachmentPaths`, that is definitely a better name

Sources/Public/Internal/SignalContext.swift
rqbacktrace: We need `attachmentPathsArray` for live reports. I tagged you where we use `attachmentPathsArray` in `BacktraceReporter`. 

`attachmentPaths` is definitely a better name, will rename.

Sources/Features/Client/Model/AttachmentsStorage.swift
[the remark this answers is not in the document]
rqbacktrace: I think I have figured out a way to unify these using a common protocol. Please see the latest version!

Sources/Features/Client/Model/AttachmentsStorage.swift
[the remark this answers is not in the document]
rqbacktrace: I think it looks much better now. Thank you for this good suggestion! Please see the updated version!

Tests/AttachmentStorageTests.swift
rqbacktrace: @krml19 @konraddysput Even if I mock my dummy file here, `AttachmentsStorage.store` function would still write a plist containing the attachment paths, just like `AttributesStorage.store` function writes a plist containing attributes. Is it still worth it to create a mock filesystem here for my dummy file? We wouldn't fully eliminate file system writes/dependencies anyway because of `AttachmentsStorage.store` internal details.

rqbacktrace:
@krml19 @konraddysput please see latest version!

Tests/AttachmentStorageTests.swift
krml19: I think it can be resolved by the dependency injection technique. In other words, one can swap and inject all the required external dependencies.

rqbacktrace:
@krml19 I have refactored `AttachmentsStorage` and `AttributesStorage` to allow injection of `ReportMetadataStorage` and updated the unit tests to use a mocked `ReportMetadataStorageMock`. Please have another look! Thank you!

Sources/Public/BacktraceCrashReporter.swift
```diff
@@ -5,6 +5,7 @@ import Backtrace_PLCrashReporter
 @objc public class BacktraceCrashReporter: NSObject {
     private let reporter: PLCrashReporter
     static private let crashName = "live_report"
+    private var copiedFileAttachments = [URL]()
```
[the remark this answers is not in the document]
rqbacktrace: Done, thank you!

konraddysput (approved):
Nothing critical on my side. Thanks a lot @rqbacktrace !

Tests/AttachmentStorageTests.swift
rqbacktrace: @krml19 @konraddysput I have injected all filesystem dependencies, please take another look!

Examples/Example-iOS/AppDelegate.swift
```diff
@@ -46,6 +52,30 @@ final class AppDelegate: UIResponder, UIApplicationDelegate {
...
+                    attributes: nil
+                )
+        let fileUrl = directoryUrl.appendingPathComponent(fileName)
+        let formatter = DateFormatter()
+        formatter.timeStyle = .medium
+        let myData = formatter.string(from: Date())
+        do {
+            try myData.write(to: fileUrl, atomically: true, encoding: .utf8)
+        } catch {
+            print("Error: \(error)")
```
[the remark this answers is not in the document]
rqbacktrace: @konraddysput apologies...I don't quite understand. Can you please give me some more details?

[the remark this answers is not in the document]
konraddysput: Right now we're printing error to stdout. Does it make sense to fail instead? Because otherwise user will assume attachment support doesn't work.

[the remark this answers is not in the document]
konraddysput: What do you think?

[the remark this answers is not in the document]
rqbacktrace: Ah good idea @konraddysput, I will make this change

Sources/Public/BacktraceClientCustomizing.swift
```diff
@@ -6,12 +6,23 @@ public typealias BacktraceClientProtocol = BacktraceReporting & BacktraceClientC
 /// Type-alias of passing attributes to library.
 public typealias Attributes = [String: Any]
 
+/// Type-alias of passing file attachments to library.
+/// Expected format: Filename, File URL bookmark
+public typealias Attachments = [String: URL]
+
+/// Type-alias of storing file attachments on disk (as a bookmark)
+/// Expected format: Filename, File URL bookmark
+public typealias Bookmarks = [String: Data]
```
krml19, line 15: Does it need to be public?

krml19 (approved):
LGTM 👍

Sources/Public/BacktraceClientCustomizing.swift
rqbacktrace: @krml19 if it's private I won't be able to use it in my `AttachmentBookmarkHandlerMock` class

PR #60, opened 2021-04-09
Remove Bluetooth listener

### LINKED ISSUE #53: BluetoothStatusListener causes App store submission to require usage description
This file imports CoreBluetooth and uses CBCentralManager, CBCentralManagerDelegate and CBCentralManagerOptionShowPowerAlertKey.

While the framework doesn't ever reference BluetoothStatusListener, its built as part of the target by default and appears to trigger Apple to ask for the following upon app store submission :

NSBluetoothPeripheralUsageDescription
NSBluetoothAlwaysUsageDescription

It would be preferable to unlink the file from the target (or remove the file if it isn't used or necessary).

krml19:
Closes #53


rqbacktrace:
Looks good to me! We just remove an unnecessary file `BluetoothStatusListener.swift`?

rqbacktrace (approved):

PR #63, opened 2021-04-19
Clean up file attachments API

rqbacktrace:
Let the user specify their file attachments as simply a list of URLs. Previously we required the user to specify their file attachments as a dictionary of String/URL where the strings are the filename. But we can extract the filename directly from the URL to provide a simpler API for the user.

### DIFF
--- a/Examples/Example-iOS/AppDelegate.swift
+++ b/Examples/Example-iOS/AppDelegate.swift
@@ -36,9 +36,12 @@ final class AppDelegate: UIResponder, UIApplicationDelegate {
         BacktraceClient.shared?.attributes = ["foo": "bar", "testing": true]
         
         let fileName = "sample.txt"
-        let fileUrl = try? createAndWriteFile(fileName)
+        guard let fileUrl = try? createAndWriteFile(fileName) else {
+            print("Could not create the file attachment")
+            return false
+        }
         var crashAttachments = Attachments()
-        crashAttachments[fileName] = fileUrl
+        crashAttachments.append(fileUrl)
         BacktraceClient.shared?.attachments = crashAttachments
 
         BacktraceClient.shared?.loggingDestinations = [BacktraceBaseDestination(level: .debug)]
--- a/Sources/Features/Client/Model/AttachmentBookmarkHandler.swift
+++ b/Sources/Features/Client/Model/AttachmentBookmarkHandler.swift
@@ -10,9 +10,8 @@ enum AttachmentBookmarkHandlerImpl: AttachmentBookmarkHandler {
         var attachmentsBookmarksDict = Bookmarks()
         for attachment in attachments {
             do {
-                let bookmark = try attachment.bookmarkData(options: URL.BookmarkCreationOptions.minimalBookmark)
-                let fileName = attachment.lastPathComponent
-                attachmentsBookmarksDict[fileName] = bookmark
+                let bookmark = try attachment.bookmarkData(options: .minimalBookmark)
+                attachmentsBookmarksDict[attachment.path] = bookmark
             } catch {
                 BacktraceLogger.error("Could not bookmark attachment file URL. Error: \(error)")
                 continue
--- a/Sources/Features/Client/Model/AttachmentBookmarkHandler.swift
+++ b/Sources/Features/Client/Model/AttachmentBookmarkHandler.swift
@@ -10,8 +10,9 @@ enum AttachmentBookmarkHandlerImpl: AttachmentBookmarkHandler {
         var attachmentsBookmarksDict = Bookmarks()
         for attachment in attachments {
             do {
-                let bookmark = try attachment.value.bookmarkData(options: URL.BookmarkCreationOptions.minimalBookmark)
-                attachmentsBookmarksDict[attachment.key] = bookmark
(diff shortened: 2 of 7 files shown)


vlussenburg (approved):

konraddysput (approved):

Examples/Example-iOS/AppDelegate.swift
```diff
@@ -36,9 +36,12 @@ final class AppDelegate: UIResponder, UIApplicationDelegate {
         BacktraceClient.shared?.attributes = ["foo": "bar", "testing": true]
         
         let fileName = "sample.txt"
-        let fileUrl = try? createAndWriteFile(fileName)
+        guard let fileUrl = try? createAndWriteFile(fileName) else {
+            print("Could not create the file attachment")
```
krml19, line 40:
```suggestion
            BacktraceLogger.error("Could not create the file attachment")
```

Sources/Features/Client/Model/AttachmentBookmarkHandler.swift
```diff
@@ -10,8 +10,9 @@ enum AttachmentBookmarkHandlerImpl: AttachmentBookmarkHandler {
         var attachmentsBookmarksDict = Bookmarks()
         for attachment in attachments {
             do {
-                let bookmark = try attachment.value.bookmarkData(options: URL.BookmarkCreationOptions.minimalBookmark)
-                attachmentsBookmarksDict[attachment.key] = bookmark
+                let bookmark = try attachment.bookmarkData(options: URL.BookmarkCreationOptions.minimalBookmark)
```
krml19, line 13: Short-hand syntax should work as well:

```suggestion
                let bookmark = try attachment.bookmarkData(options: .minimalBookmark)
```

Sources/Features/Client/Model/AttachmentBookmarkHandler.swift
```diff
@@ -10,8 +10,9 @@ enum AttachmentBookmarkHandlerImpl: AttachmentBookmarkHandler {
...
+                let fileName = attachment.lastPathComponent
+                attachmentsBookmarksDict[fileName] = bookmark
```
krml19, line 15: Please consider using the entire path for a key. Otherwise, some keys might get overridden, e.g.:
 - `foo/attachment.log`
 - `bar/attachment.log`

krml19 (commented):
Looks good to me. Left some comment for possible improvements 👍

Examples/Example-iOS/AppDelegate.swift
rqbacktrace: I don't think we have access to `BacktraceLogger` here. We use a normal `print` elsewhere in this class too.

Sources/Features/Client/Model/AttachmentBookmarkHandler.swift
rqbacktrace: Will do, thank you!

Sources/Features/Client/Model/AttachmentBookmarkHandler.swift
rqbacktrace: That's a good point...however currently we back up the file attachments to a different directory based on just their filename. We would also have to go back there and disambiguate the copied file attachments as well. This is definitely something we can improve and I will create a ticket for this internally. Until then, we will have to make it clear to users we cannot send two file attachments with the same filename using this method.

Examples/Example-iOS/AppDelegate.swift
krml19: I'd recommend removing all print statements from the entire codebase and replacing them with proper log levels. This way, we can control how many events are visible for our end users. Please note that the `Could not create the file attachment` log will be added to the application without any other context, so developers might get confused.

Sources/Features/Client/Model/AttachmentBookmarkHandler.swift
rqbacktrace: But until then it's still better to make this change, I will do it!

Examples/Example-iOS/AppDelegate.swift
rqbacktrace: I can do this but I will have to make BacktraceLogger functions public @krml19

krml19: @rqbacktrace There is no need for that. Those methods are used within other methods' scopes, so it should work just fine.

Sources/Features/Client/Model/AttachmentBookmarkHandler.swift
krml19: Ok, so I'll mark it as resolved then.

Examples/Example-iOS/AppDelegate.swift
rqbacktrace: @krml19 it doesn't build unless I make `BacktraceLogger` functions public. I get `info` is inaccessable due to `internal` protection level.

rqbacktrace: Based on internal discussion with @krml19 we will not make this change

krml19: It's just an example project. Please ignore my previous comments

krml19 (approved):
LGTM 👍

PR #69, opened 2021-11-09
Make hostname optional since resolving hostname will cause Local Network permissions dialog to appear

rqbacktrace:


rqbacktrace:
I think the CI is a little shaky, all tests pass locally using fastlane but I had to re-run this test

konraddysput (approved):

vlussenburg (approved):

PR #72, opened 2021-12-16
Feature/error-free metrics

rqbacktrace:
This adds error-free metrics to backtrace-cocoa. This allows app developers to answer the following questions:

- How many unique users of my app (as measured by device ID) are experiencing errors/crashes?
- How many sessions of my app (as measured by application launches) are experiencing errors/crashes?

If a user experiences an error/crash using the app, it will be counted against crash-free (error-free) users.
Similarly, if an app session (new application launch) experiences an error/crash, it will be counted against crash-free (error-free) application launches.

For E2E testing:
Enable metrics using a Backtrace Universe which has error-free metrics enabled and provide your normal crash submission token. You should get 200 OK responses to requests to the error-free metrics server.
After about 30 minutes you should be able to click the Overview tab in the web UI and view crash-free (error-free) users and application launches.

Future work:
- Configuration
- Add per metrics event attributes
- Auto-send metrics events every X minutes
- Retry metrics send on 50X response code


vlussenburg (approved):

Sources/Public/BacktraceCredentials.swift
```diff
@@ -28,3 +28,92 @@ import Foundation
...
+                let tokenStartIndex = urlString.index(after: urlString.index(tokenEndIndex, offsetBy: -(tokenLength - 1)))
+
+                return String(urlString[tokenStartIndex...tokenEndIndex])
+            } else {
+                guard let tokenQueryParamRange = urlString.range(of: tokenQueryParam) else {
+                    throw BacktraceUrlParsingError.invalidInput(urlString)
+                }
+
+                let tokenStartIndex = tokenQueryParamRange.upperBound
+                print(urlString[tokenStartIndex...urlString.index(before: urlString.endIndex)])
```
krml19, line 110: Prints should never be used in the production codebase. Please remove it or use a proper logger if this log is useful for debugging.

krml19, line 110: Instead of doing manual parsing, I'd suggest using `URLComponents` to extract URL components, e.g. URL query parameters.

Sources/Features/Metrics/Model/UniqueEventsPayload.swift
```diff
@@ -0,0 +1,14 @@
+import Foundation
+
+class UniqueEventsPayload: Payload<UniqueEvent> {
```
krml19, line 3: Can it be `final`?

Sources/Features/Metrics/Model/SummedEventsPayload.swift
```diff
@@ -0,0 +1,14 @@
+import Foundation
+
+class SummedEventsPayload: Payload<SummedEvent> {
```
krml19, line 3: Can it be `final`?

Sources/Features/Metrics/Model/MetricsRequest.swift
```diff
@@ -0,0 +1,38 @@
...
+
+extension MetricsRequest {
+    static func writeMetricsRequest<T: Event>(urlRequest: URLRequest, payload: Payload<T>) throws -> URLRequest {
+        let jsonEncoder = JSONEncoder()
+        let body = try jsonEncoder.encode(payload)
+
+        var metricsRequest = urlRequest
+        metricsRequest.setValue("application/json", forHTTPHeaderField: "Content-Type")
+        metricsRequest.httpBody = body
+    //    metricsRequest.setValue("\(body.count)", forHTTPHeaderField: "Content-Length")
```
krml19, line 34: Please remove commented out code

Sources/Features/Metrics/Model/Event.swift
```diff
@@ -0,0 +1,26 @@
+import Foundation
+
+class Event: Encodable {
```
krml19, line 3: Please consider using immutable data types when there is no need to mutable the properties of a given object. Also, I'd suggest using `struct`s for data objects rather than `class`es. `struct`s are passed by value, so they are safe in a multithreading environment. See more: https://docs.swift.org/swift-book/LanguageGuide/ClassesAndStructures.html

Sources/Features/Metrics/Model/Event.swift
```diff
@@ -0,0 +1,26 @@
...
+
+    private enum CodingKeys: String, CodingKey {
+        case timestamp, attributes
+    }
+
+    func encode(to encoder: Encoder) throws {
+        var container = encoder.container(keyedBy: CodingKeys.self)
+        try container.encode(timestamp, forKey: .timestamp)
+        try container.encode(attributes, forKey: .attributes)
+    }
```
krml19, line 25: `encode` method can be automatically synthesized when coding keys are identical to the properties names

Sources/Features/Metrics/Model/BacktraceMetricsSender.swift
```diff
@@ -0,0 +1,91 @@
...
+        }
+        return url
+    }
+
+    private func handleSummedEventsResult(result: BacktraceMetricsResult) {
+        metricsContainer.clearSummedEvents()
+    }
+
+    private func handleUniqueEventsResult(result: BacktraceMetricsResult) {
+
```
krml19, line 89: Should it be empty?

Sources/Features/Metrics/Model/BacktraceMetricsSender.swift
```diff
@@ -0,0 +1,91 @@
+import Foundation
+
+final class BacktraceMetricsSender {
+
+    private let api: BacktraceApi
+    private let metricsContainer: BacktraceMetricsContainer
+    private let settings: BacktraceMetricsSettings
+
+    static let defaultBaseUrl = "https://events.backtrace.io/api"
```
krml19, line 9: I'd suggest storing all constant values in one place, so they can be easily updated/maintained

Sources/Features/Metrics/Model/SummedEvent.swift
```diff
@@ -0,0 +1,21 @@
+import Foundation
+
+final class SummedEvent: Event {
+
+    internal(set) public var name: String
```
krml19, line 5: It seems to be wrong. The class is `internal` so setter and getter are `internal` as well.

Sources/Features/Metrics/Model/UniqueEvent.swift
```diff
@@ -0,0 +1,22 @@
+import Foundation
+
+final class UniqueEvent: Event {
+
+    // Backtrace API requires unique event name to be a JSON array
+    internal(set) public var name = [String]()
```
krml19, line 6: It seems to be wrong. The class is `internal` so setter and getter are `internal` as well.

Sources/Features/Metrics/Model/BacktraceMetricsSender.swift
```diff
@@ -0,0 +1,91 @@
...
+        } catch {
+            BacktraceLogger.error(error)
+        }
+    }
+
+    func getSubmissionUrl(urlPrefix: MetricsUrlPrefix) throws -> URL {
+        let token = try api.credentials.getSubmissionToken()
+        let universe = try api.credentials.getUniverseName()
+
+        let urlString = self.baseUrl + "/" + urlPrefix.description + "/submit?token=" + token + "&universe=" + universe
```
krml19, line 76: It's fine but I'd suggest using `URLComponents` to create complex URLs. It'll make it easier to understand.

Sources/Features/Metrics/Model/BacktraceMetricsSender.swift
konraddysput: @rqbacktrace shouldn't we update the timestamp here?

Sources/Features/Network/BacktraceApi.swift
```diff
@@ -56,3 +61,42 @@ extension BacktraceApi: BacktraceApiProtocol {
...
+
+            // get result
+            BacktraceLogger.debug("Received HTTP response from metrics API: \(httpResponse)")
+            let result = httpResponse.result()
+            metricsDelegate?.serverDidRespond?(result)
+
+            return result
+        } catch {
+            BacktraceLogger.error("Connection for \(payload) failed with error: \(error)")
+            metricsDelegate?.connectionDidFail?(error)
```
konraddysput, line 98: Shouldn't we try a retry here? (exponential backoff etc)

konraddysput (approved):
I think I have few questions regarding retry settings, but depends on priorities we can apply more complex retry settings or not.

README.md
```diff
@@ -417,6 +418,27 @@ UserDefaults.standard.register(defaults: ["NSApplicationCrashOnExceptions": true
 ```
 but it crashes your app if you don't use `@try ... @catch`.
 
+## Error-free metrics <a name="documentation-metrics"><a/>
```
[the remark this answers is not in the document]
rqbacktrace: Thank you!

Sources/Features/Metrics/Model/BacktraceMetricsHttpResponse.swift
```diff
@@ -0,0 +1,16 @@
+import Foundation
+
+struct BacktraceMetricsHttpResponse {
+
+    let statusCode: Int
+
+    init(httpResponse: HTTPURLResponse, responseData: Data?) {
```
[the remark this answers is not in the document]
rqbacktrace: Will remove this, thank you!

Sources/Public/BacktraceCredentials.swift
rqbacktrace: Will remove this, thank you!

Sources/Features/Metrics/Model/UniqueEventsPayload.swift
rqbacktrace: Will do!

Sources/Features/Metrics/Model/SummedEventsPayload.swift
rqbacktrace: Will do!

Sources/Features/Metrics/Model/MetricsRequest.swift
rqbacktrace: Will do!

Sources/Features/Metrics/Model/BacktraceMetricsSender.swift
rqbacktrace: That's right, thank you @konraddysput! Will fix!

Sources/Features/Metrics/Model/BacktraceMetricsSender.swift
rqbacktrace: Will do, thank you!

Sources/Features/Metrics/Model/UniqueEvent.swift
rqbacktrace: Will remove, thank you!

Sources/Features/Network/BacktraceApi.swift
rqbacktrace: We are not doing retry for V1 of iOS metrics, @vlussenburg

Sources/Features/Metrics/Model/Event.swift
rqbacktrace: Will do, thank you!

rqbacktrace: Will do, thank you!

Sources/Public/BacktraceCredentials.swift
rqbacktrace: I think it looks much better after making your suggested change, thank you!

Sources/Features/Metrics/Model/BacktraceMetricsSender.swift
rqbacktrace: I understand and agree in general but I think in this case it becomes more difficult to understand because of all the error handling:

```swift
        let token = try api.credentials.getSubmissionToken()
        let universe = try api.credentials.getUniverseName()

        guard let baseUrl = URL(string: baseUrlString) else {
            throw BacktraceUrlParsingError.invalidInput(baseUrlString)
        }

        guard var components = URLComponents(url: baseUrl, resolvingAgainstBaseURL: true) else {
            throw BacktraceUrlParsingError.invalidInput(baseUrl.debugDescription)
        }

        components.path = urlPrefix.description + "/submit"
        components.queryItems = [
            URLQueryItem(name: "token", value: token),
            URLQueryItem(name: "universe", value: universe)
        ]

        guard let url = components.url else {
            throw BacktraceUrlParsingError.invalidInput(baseUrl.debugDescription)
        }

        return url
```

rqbacktrace:
@krml19 I've addressed your review feedback, can you please take another look? Thank you!

Sources/Features/Metrics/Model/BacktraceMetricsSender.swift
krml19: Personally, I am ok with those changes. You're right that there is a bit more code for error handling but it gives a nice overview of what's happening. It's up to you to decide and I'm not going to block PR for this reason 😉

Sources/Features/Metrics/Model/Event.swift
```diff
@@ -0,0 +1,21 @@
...
+protocol Event: Encodable {
+
+    var timestamp: Int64 { get set }
+    var attributes: DecodableAttributes { get set }
+}
+
+extension Event {
+    func initialTimestamp() -> Int64 {
+        return Date().currentTimeSeconds()
+    }
```
krml19, line 12: Just a suggestion to make more testable: 

```suggestion
    static func initialTimestamp(_ date: Date = Date()) -> Int64 {
        return date.currentTimeSeconds()
    }
```

Sources/Features/Metrics/Model/Event.swift
```diff
@@ -0,0 +1,21 @@
...
+
+    func initialAttributes() -> DecodableAttributes {
+        let attributesProvider = AttributesProvider()
+        let localAttributes = attributesProvider.allAttributes
+        let localAttributesConverted: [String: String] = localAttributes.compactMapValues { "\($0)" }
+
+        return localAttributesConverted
+    }
```
krml19, line 20:
```suggestion
    static func initialAttributes(_ attributesProvider: AttributesProvider = AttributesProvider()) -> DecodableAttributes {
        let localAttributes = attributesProvider.allAttributes
        let localAttributesConverted: [String: String] = localAttributes.compactMapValues { "\($0)" }

        return localAttributesConverted
    }
```

Sources/Features/Metrics/Model/MetricsRequest.swift
```diff
@@ -0,0 +1,50 @@
...
+        return urlRequest
+    }
+}
+
+// From: https://gist.github.com/sourleangchhean168/f1a663c8524936af35221f410b588677
+private extension Data {
+    var prettyPrintedJSONString: NSString? {
+        guard let object = try? JSONSerialization.jsonObject(with: self, options: []),
+              let data = try? JSONSerialization.data(withJSONObject: object, options: [.prettyPrinted]),
+              let prettyPrintedString = NSString(data: data, encoding: String.Encoding.utf8.rawValue) else { return nil }
```
krml19, line 31:
```suggestion
              let prettyPrintedString = String(data: data, encoding: .utf8) else { return nil }
```

Sources/Features/Metrics/Model/Payload.swift
```diff
@@ -0,0 +1,33 @@
+import Foundation
+
+struct EventsMetadata: Encodable {
+
+    var droppedEvents = 0
```
krml19, line 5: Should it be mutable? If so, it's fine. Otherwise, please use the `let` keyword. It's minor but it using immutable data types creates fewer bugs, so I'd prefer immutable properties in the first place and make it mutable if and only if it needs to be.

Sources/Features/Metrics/Model/SummedEvent.swift
```diff
@@ -0,0 +1,26 @@
+import Foundation
+
+struct SummedEvent: Event {
+
+    var timestamp = Int64()
+    var attributes = DecodableAttributes()
```
krml19, line 6: Those properties are always set in the initializer

```suggestion
    let timestamp: Int64
    let attributes: DecodableAttributes
```

Sources/Features/Metrics/Model/SummedEventsPayload.swift
```diff
@@ -0,0 +1,29 @@
+import Foundation
+
+struct SummedEventsPayload: Payload {
+
+    typealias Event = SummedEvent
+
+    var applicationName = String()
+    var applicationVersion = String()
+    var metadata = EventsMetadata()
+    var events = [SummedEvent]()
```
krml19, line 10: Same as above

Sources/Features/Metrics/Model/UniqueEvent.swift
```diff
@@ -0,0 +1,27 @@
+import Foundation
+
+struct UniqueEvent: Event {
+
+    var timestamp = Int64()
```
krml19, line 5: Same as above

Sources/Features/Metrics/Model/UniqueEventsPayload.swift
```diff
@@ -0,0 +1,29 @@
+import Foundation
+
+struct UniqueEventsPayload: Payload {
+
+    typealias Event = UniqueEvent
+
+    var applicationName = String()
```
krml19, line 7: Same as above

Sources/Features/Metrics/Model/BacktraceMetricsSender.swift
rqbacktrace: Good idea, will do, thank you!

Sources/Features/Metrics/Model/Event.swift
rqbacktrace: Good idea, thank you!

Sources/Features/Metrics/Model/Event.swift
rqbacktrace: Will do!

Sources/Features/Metrics/Model/MetricsRequest.swift
rqbacktrace: Will do!

Sources/Features/Metrics/Model/Payload.swift
rqbacktrace: This one does need to be mutable. In the future droppedEvents will increment on retry

Sources/Features/Metrics/Model/SummedEvent.swift
rqbacktrace: Will do, thank you!

Sources/Features/Metrics/Model/SummedEventsPayload.swift
rqbacktrace: Will fix!

Sources/Features/Metrics/Model/BacktraceMetricsSender.swift
rqbacktrace: I will use the URLComponents method, thank you!

Sources/Features/Metrics/Model/SummedEvent.swift
rqbacktrace: I will defer initialization to the initializer but in this case as well in the future timestamp and attributes needs to be mutable for retry logic

Sources/Features/Metrics/Model/UniqueEvent.swift
rqbacktrace: Will do!

Sources/Features/Metrics/Model/UniqueEventsPayload.swift
rqbacktrace: Will do!

Sources/Features/Metrics/Model/UniqueEvent.swift
rqbacktrace: Will do!

Sources/Features/Metrics/Model/SummedEventsPayload.swift
rqbacktrace: Will do!

rqbacktrace:
@krml19 I've addressed the second round of review feedback, can you please take another look?

krml19 (approved):

PR #73, opened 2022-05-04
Disabled OOM reports

MinPD126:
Disabled OOM reports for 15.3.1+ for backtrace-cocoa (so invalid OOM reports don't crash it)

### DIFF
--- a/Sources/Features/Client/BacktraceReporter.swift
+++ b/Sources/Features/Client/BacktraceReporter.swift
@@ -32,7 +32,7 @@ final class BacktraceReporter {
         let attributesProvider = AttributesProvider(reportHostName: dbSettings.reportHostName)
         self.attributesProvider = attributesProvider
         if #available(iOS 15.3.1, *) {
-            BacktraceLogger.debug("The OOM support is disadbled for this version of iOS. Skipping OOM check.")
+            BacktraceLogger.debug("The OOM support is disabled for this version of iOS. Skipping OOM check.")
             oomSupport = false
         }
         if(oomSupport == true){
--- a/Sources/Features/Client/BacktraceReporter.swift
+++ b/Sources/Features/Client/BacktraceReporter.swift
@@ -6,8 +6,9 @@ final class BacktraceReporter {
     private(set) var api: BacktraceApi
     private let watcher: BacktraceWatcher<PersistentRepository<BacktraceReport>>
     private(set) var attributesProvider: SignalContext
-    private(set) var backtraceOomWatcher: BacktraceOomWatcher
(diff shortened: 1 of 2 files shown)


Sources/Features/Client/BacktraceReporter.swift
```diff
@@ -30,11 +31,17 @@ final class BacktraceReporter {
         self.repository = try PersistentRepository<BacktraceReport>(settings: dbSettings)
         let attributesProvider = AttributesProvider(reportHostName: dbSettings.reportHostName)
         self.attributesProvider = attributesProvider
-        self.backtraceOomWatcher = BacktraceOomWatcher(
-            repository: self.repository,
-            crashReporter: self.reporter,
-            attributes: attributesProvider,
-            backtraceApi: self.api)
+        if #available(iOS 15.3.1, *) {
+            BacktraceLogger.debug("The OOM support is disadbled for this version of iOS. Skipping OOM check.")
```
vlussenburg, line 35: typo disadbled => disabled

Sources/Features/Client/BacktraceReporter.swift
```diff
@@ -30,11 +31,17 @@ final class BacktraceReporter {
...
+            oomSupport = false
+        }
+        if(oomSupport == true){
```
vlussenburg, line 38: - if style seems to be without parentheses in this app ie `if self.reporter.oomSupport` vs `if(oomSupport == true)`
- also, the `== true` seems superfluous here
- could the `oomSupport` bool be named better? We now have `detectOom` and `oomSupport` which seems like the same property. Would it be better to force `self.configuration.detectOom` to be false instead of introduced another boolean?

Sources/Features/Client/BacktraceReporter.swift
```diff
@@ -6,8 +6,9 @@ final class BacktraceReporter {
     private(set) var api: BacktraceApi
     private let watcher: BacktraceWatcher<PersistentRepository<BacktraceReport>>
     private(set) var attributesProvider: SignalContext
-    private(set) var backtraceOomWatcher: BacktraceOomWatcher
+    private(set) var backtraceOomWatcher: BacktraceOomWatcher!
     let repository: PersistentRepository<BacktraceReport>
+    var oomSupport : Bool = true
```
vlussenburg, line 11: Should it be inited with true? Or `self.configuration.detectOom`? Or maybe we can avoid this var altogether.

Sources/Features/Client/BacktraceReporter.swift
MinPD126: These have been implemented

vlussenburg (approved):

vlussenburg:
Approved for release.

PR #75, opened 2022-05-27
Feature/setup swiftPM

MinPD126:
Integration of backtrace swift package manager for iOS, macOS and tvOS.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version: 5.6
+// swift-tools-version: 5.4
 // The swift-tools-version declares the minimum version of Swift required to build this package.
 
 import PackageDescription
--- a/Package.swift
+++ b/Package.swift
@@ -27,7 +27,7 @@ let package = Package(
                 .product(name: "CrashReporter", package: "plcrashreporter")
             ],
             path: "Sources",
-            resources: [.copy("Features/Resources/Model.xcdatamodeld")]
+            resources: [.process("Features/Resources/Model.xcdatamodeld")]
         ),
         .testTarget(
             name: "BacktraceTests",
--- a/Package.swift
+++ b/Package.swift
@@ -26,7 +26,8 @@ let package = Package(
             dependencies: [
                 .product(name: "CrashReporter", package: "plcrashreporter")
             ],
-            path: "Sources"
+            path: "Sources",
+            resources: [.copy("Features/Resources/Model.xcdatamodeld")]
         ),
         .testTarget(
             name: "BacktraceTests",
--- a/Package.swift
+++ b/Package.swift
@@ -6,7 +6,9 @@ import PackageDescription
 let package = Package(
     name: "Backtrace",
     platforms: [
-        .iOS(.v10)
+        .iOS(.v10),
+        .macOS(.v10_10),
+        .tvOS(.v12)
     ],
     products: [
         .library(
--- a/Package.swift
+++ b/Package.swift
@@ -1,0 +1,28 @@
+// swift-tools-version: 5.6
+// The swift-tools-version declares the minimum version of Swift required to build this package.
+
+import PackageDescription
+
+let package = Package(
+    name: "Backtrace",
+    platforms: [
+        .iOS(.v10)
+    ],
+    products: [
+        .library(
+            name: "Backtrace",
+            targets: ["Backtrace"]),
+    ],
+    dependencies: [
+        .package(url: "https://github.com/microsoft/plcrashreporter.git", from: "1.10.0"),
+    ],
+    targets: [
+        .target(
+            name: "Backtrace",
+            dependencies: [
+                .product(name: "CrashReporter", package: "plcrashreporter")
+            ],
+            path: "Sources"
+        )
+    ]
+)
--- a/Package.swift
+++ b/Package.swift
@@ -26,8 +26,7 @@ let package = Package(
             dependencies: [
                 .product(name: "CrashReporter", package: "plcrashreporter")
             ],
-            path: "Sources",
-            resources: [.process("Features/Resources/Model.xcdatamodeld")]
+            path: "Sources"
         ),
         .testTarget(
             name: "BacktraceTests",
--- a/Package.swift
+++ b/Package.swift
@@ -17,6 +17,8 @@ let package = Package(
     ],
     dependencies: [
         .package(url: "https://github.com/microsoft/plcrashreporter.git", from: "1.10.0"),
+        .package(url: "https://github.com/Quick/Nimble.git", from: "9.1.1"),
+        .package(url: "https://github.com/Quick/Quick.git", from: "4.0.0")
     ],
     targets: [
         .target(
@@ -25,6 +27,18 @@ let package = Package(
                 .product(name: "CrashReporter", package: "plcrashreporter")
             ],
             path: "Sources"
+        ),
+        .testTarget(
+            name: "BacktraceTests",
+            dependencies: [
+                "Backtrace",
+                .product(name: "CrashReporter", package: "plcrashreporter"),
+                .product(name: "Nimble", package: "Nimble"),
+                .product(name: "Quick", package: "Quick")
+            ],
+            path: "Tests",
+            resources: [.process("Resources/test.txt")]
         )
     ]
 )
+
--- a/Sources/Features/Attributes/DefaultAttributes.swift
+++ b/Sources/Features/Attributes/DefaultAttributes.swift
@@ -1,6 +1,7 @@
 // swiftlint:disable type_name
 import Foundation
 import CoreLocation
+import UIKit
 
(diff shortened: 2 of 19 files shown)


Package.swift
```diff
@@ -0,0 +1,30 @@
...
+        .macOS(.v10_10),
+        .tvOS(.v12)
+    ],
+    products: [
+        .library(
+            name: "Backtrace",
+            targets: ["Backtrace"]),
+    ],
+    dependencies: [
+        .package(url: "https://github.com/microsoft/plcrashreporter.git", from: "1.10.0"),
```
vlussenburg, line 19: Is it possible to create an initial version of this that depends on our [PLCR fork](https://github.com/backtrace-labs/plcrashreporter)? Probably not, since our fork is not SwiftPM enabled, right?

Sources/Features/Attributes/DefaultAttributes.swift
```diff
@@ -203,7 +208,7 @@ struct LibInfo: AttributesSource {
     var immutable: [String: Any?] {
         return ["guid": LibInfo.guid(store: UserDefaultsStore.self).uuidString,
                 "lang.name": LibInfo.applicationLangName,
-                "lang.version": BacktraceVersionNumber]
+                "lang.version": Bundle.main.releaseVersionNumber]
```
vlussenburg, line 211: did you test this for MacOS and iOS that it yields the right version number? Do/can we unit test this?

vlussenburg: On second thought, I think this populated incorrectly. See for Android: https://github.com/backtrace-labs/backtrace-android/blob/2f57da44a071d3f382b6a911db995d8d09c0e19f/backtrace-library/src/main/java/backtraceio/library/models/BacktraceData.java#L180 and Unity https://github.com/backtrace-labs/backtrace-unity/blob/f42386b496b741504b6223190bef790419e63874/Runtime/Model/BacktraceData.cs#L144

Although it's a bit of a mess, I think this property is meant to be system version. Can we add the macOS/iOS version here instead? 

Can you add this `Bundle.main.releaseVersionNumber` to `backtrace.version` instead? This is the field we use for the library's version (NOT the version of the app using our library)

Sources/Features/Attributes/DefaultAttributes.swift
```diff
@@ -232,7 +237,7 @@ struct MetricsInfo: AttributesSource {
 
     var immutable: [String: Any?] {
         return MetricsInfo.isMetricsEnabled ?
-            ["application.version": Backtrace.applicationVersion,
+            ["application.version": Bundle.main.releaseVersionNumber,
```
vlussenburg, line 240: This should be the version of the client app using our library, so I think this is right. We should verify this though.

Sources/Features/Repository/Model/BacktraceReport.swift
```diff
@@ -1,5 +1,9 @@
 import Foundation
+#if canImport(CrashReporter)
+import CrashReporter
+#elseif canImport(Backtrace_PLCrashReporter)
 import Backtrace_PLCrashReporter
```
vlussenburg, line 5: Is this still here for backwards compatibility if someone builds our lib using Cocoa-Pods?

Sources/Public/BacktraceCrashReporter.swift
```diff
@@ -48,7 +53,11 @@ extension BacktraceCrashReporter: CrashReporting {
                             attributes: Attributes,
                             attachmentPaths: [String] = []) throws -> BacktraceReport {
 
+#if SWIFT_PACKAGE
```
vlussenburg, line 56: Why is there an ifdef here, why does it need to be different for SWIFT_PACKAGE vs other build options?

Package.swift
```diff
@@ -0,0 +1,30 @@
+// swift-tools-version: 5.6
+// The swift-tools-version declares the minimum version of Swift required to build this package.
+
+import PackageDescription
+
+let package = Package(
+    name: "Backtrace",
+    platforms: [
+        .iOS(.v10),
```
vlussenburg, line 9: Where did you get these minimum versions from? They look good, but just checking

MinPD126: We have taken the refrence of exisitng targets like Example-iOS, Backtrace-iOS etc.

Sources/Public/BacktraceCrashReporter.swift
MinPD126: With Swift package manager we are using https://github.com/microsoft/plcrashreporter.git which is orignal branch with latest code but with framework we are still using Backtrash_CrashReporter which is significantly old one. so as per the build oprion we picked implementation.

Sources/Features/Repository/Model/BacktraceReport.swift
MinPD126: For Cocoa pod its will works well, we maintian the backward compatiblity.

Package.swift
MinPD126: Yes its not possible as its not enabled with SwiftPM and out of date as well.

Sources/Features/Attributes/DefaultAttributes.swift
MinPD126: backtrace.version:Do we want Backtrace version and how are we extracting it .

vlussenburg: 1) This version should never have been set in `lang.version`. This should be populate by iOS or macOS version.

So we want:
- `agent` is `backtrace-macOS`, `backtrace-tvOS` or `backtrace-iOS` (basically `backtrace-` + getSysname)
- `agent.version` is version of our library (see further below)
- `lang.name` is macOS, tvOS or iOS (same as `uname.sysname`) 
- `lang.version` is the macOs/tvOS/iOS version

Why it's not resolving: it seems to be a cocoa feature:
https://github.com/backtrace-labs/backtrace-cocoa/blob/6cb3b4b1538369320d750d7fd52776e614828579/Backtrace-iOS/Backtrace.h
https://stackoverflow.com/questions/31418303/get-cocoa-touch-framework-project-version-string-in-swift

You would have to find a way to resolve the library version for both cocoa and SwiftPM, I think this will do that:
https://stackoverflow.com/questions/24538215/get-version-number-of-ios-universal-framework-in-client

Sources/Features/Repository/Model/BacktraceReport.swift
vlussenburg: As we discussed, we do not want to release a version where we rely on a different versions of plcrashreporter for SwiftPM vs Cocoapods.

Please change the Podfile to use the microsoft official cocoa pod:
https://cocoapods.org/pods/PLCrashReporter

Sources/Public/BacktraceCrashReporter.swift
vlussenburg: This should be removed as per the previous comment,

Package.swift
vlussenburg: Can you please double check that there version are indeed supported by the latest version of PLCrashreporter?

vlussenburg:
I'm getting the following error:
```
package at 'https://github.com/backtrace-labs/backtrace-cocoa/' @ 17961928ff3565d8e56c4fd590ef457562141e86 is using Swift tools version 5.6.0 but the installed version is 5.5.0 in https://github.com/backtrace-labs/backtrace-cocoa/
```

Suggesting to put the lowest Swift version we support in `swift-tools-version` instead of 5.6 (latest)

vlussenburg:
Getting compile errors on my test project (running from vlussenburg-patch-1 with modified swift-tools-version)
[Build BacktraceDemoApp_2022-05-31T14-01-38.txt](https://github.com/backtrace-labs/backtrace-cocoa/files/8804899/Build.BacktraceDemoApp_2022-05-31T14-01-38.txt)

Package.swift
MinPD126: Its been verified again.

Following version supported by PLCrashreporter latest.
platforms: [
        .iOS(.v9),
        .macOS(.v10_10),
        .tvOS(.v9)
    ],

ohayon:
Hi folks, I am curious if there are thoughts on reviving this PR or any other work towards supporting SPM in the near future?

rick-bt:
Hi David - The priority for SPM support got solidly bumped up recently. We will be working on it early this year.

ohayon:
That's great news, thanks so much @rick-bt for the prompt reply. 

Do you think this PR is the best place to follow along with that work or is there a better venue to track to make sure I see when SPM support is available?

rick-bt:
That's not clear at the moment, @ohayon. A different engineer will be assigned to this, and they will determine whether or not it is appropriate to persist this PR. If not, I will do my best to link here to the new one.

rick-bt:
So, @ohayon - not a moment too soon? :)

PR #76, opened 2022-06-21
Feature/integrate breadcrumbs

vlussenburg:
# What this PR adds:

## Breadcrumbs
- Log common iOS events as breadcrumbs to give developers more debugging context
- Allow developers to add their own breadcrumbs in native or managed code to give more debugging context
- Send breadcrumb logs with managed and native reports
- Automatically add memory, battery, orientation iOS events to breadcrumbs

# To do:
- [x] Unit tests
- [x] Use a [highly performant file structure](https://github.com/linkedin/cassette/blob/master/Cassette/CASQueueFile.m) instead of read-write full file for each breadcrumb
- [x] Documentation changes (README.md)
- [x] Test report

# Test report V1 (MVP)
This is V1 of Backtrace iOS Breadcrumbs implementation

## Test Platform
- iOS version 10 and above

## Test Devices
-  iPhone 13 
    iOS 15.5
-  iPhone 12 mini 
    iOS 15.3
-  iPhone SE (2nd generation)
    iOS 14.8
-  iPhone 11 
    iOS 14.8

## Testing strategy
1. Enable breadcrumb
1. Add breadcrumbs manually
1. Breadcrumbs report (native + managed) will submit along with crash reports automatically which reflects on the console.

### Native report screenshot
image

### Managed report screenshot
image
Screen Shot 2022-07-11 at 1 20 59 PM


Sources/Public/BacktraceQueueFileHelper.swift
```diff
@@ -0,0 +1,98 @@
...
+
+extension BacktraceQueueFileHelper {
+    
+    func getSavedContent() -> [[String:Any]]? {
+        do {
+            let fileURL = URL(fileURLWithPath: breadcrumbLogDirectory)
+            let content = try String(contentsOf: fileURL, encoding: .utf8)
+            return convertStringIntoInfo(content)
+        } catch {
+            return nil
```
vlussenburg, line 76: Shouldn't we log the exception so we know why the content couldn't be retrieved?

Sources/Public/BacktraceQueueFileHelper.swift
```diff
@@ -0,0 +1,98 @@
...
+        }
+    }
+    
+    func convertStringIntoInfo(_ text: String) -> [[String:Any]]? {
+        if let data = text.data(using: .utf8) {
+            do {
+                return try JSONSerialization.jsonObject(with: data, options: []) as? [[String: Any]]
+            } catch {
+                return nil
```
vlussenburg, line 85: Shouldn't we log the exception so we know why the content couldn't be converted?

Sources/Public/BacktraceQueueFileHelper.swift
```diff
@@ -0,0 +1,98 @@
...
+                if fullContent.count == 0 {
+                    fullContent = text
+                } else {
+                    fullContent.append("\n\n\(text)")
+                }
+                try fullContent.write(to: fileURL, atomically: true, encoding: .utf8)
+                return true
+            }
+            catch {
+                BacktraceLogger.warning("\(error.localizedDescription) \nWhen ading breadcrumbs")
```
vlussenburg, line 48: typo `ading`

Sources/Public/BacktraceBreadcrumb.swift
```diff
@@ -0,0 +1,150 @@
...
+        }
+        guard let lastBreadcrumbId = getCurrentBreadcrumbId else {
+            return
+        }
+        do {
+            let fileURLPath = try breadcrumbLogPath()
+            report.attachmentPaths.append(fileURLPath)
+            report.attributes["breadcrumbs.lastId"] = lastBreadcrumbId
+        } catch {
+            BacktraceLogger.warning("\(error.localizedDescription) \nWhen process breadcrumbs report")
```
vlussenburg, line 147: Typo `process` should be `processing`

Sources/Public/BacktraceQueueFileHelper.swift
```diff
@@ -0,0 +1,98 @@
...
+    
+        if let text = convertInfoIntoString(info) {
+            let textBytes = Data(text.utf8)
+            if textBytes.count > 4096 {
+                BacktraceLogger.error("We should not have a breadcrumb this big, this is a bug!")
+                return false
+            }
+            
+            let fileURL = URL(fileURLWithPath: breadcrumbLogDirectory)
+            let content = try? String(contentsOf: fileURL, encoding: .utf8)
```
vlussenburg, line 29: Concern: performance impact of reading then write again.

Other implementation have either appended to the end of the file until the file got too big, which saves time on reading and re-writing the entire file.

vlussenburg: Another question: can the writing be done asynchronous? Is this implementation thread safe?

Sources/Public/BacktraceBreadcrumbsLogManager.swift
```diff
@@ -0,0 +1,52 @@
...
+        self.backtraceQueueFileHelper = BacktraceQueueFileHelper(breadcrumbLogPath, maxQueueFileSizeBytes: maxQueueFileSizeBytes)
+    }
+    
+    public func addBreadcrumb(_ message: String,
+                              attributes:[String:Any]? = nil,
+                              type: BacktraceBreadcrumbType,
+                              level: BacktraceBreadcrumbLevel) -> Bool {
+        let time = Date().millisecondsSince1970
+        
+        var info: [String: Any] = ["timestamp": time,
```
vlussenburg, line 20: Would consider a better name than `info`. It's probably `breadcrumb`?

Sources/Public/BacktraceBreadcrumbsLogManager.swift
```diff
@@ -0,0 +1,52 @@
...
+        if let attributes = attributes, attributes.keys.count > 0 {
+            var attribInfo: [String: Any] = [String: Any]()
+            for attribute in attributes {
+                attribInfo[attribute.key] = attribute.value
+            }
+            info["attributes"] = attribInfo
+        }
+        
+        if let result = backtraceQueueFileHelper?.addInfo(info), result == true {
+            breadcrumbId += 1
```
vlussenburg, line 34: I would auto increment always, not just when successful

Sources/Public/BacktraceBreadcrumbsLogManager.swift
```diff
@@ -0,0 +1,52 @@
+import Foundation
+
+@objc public class BacktraceBreadcrumbsLogManager: NSObject {
+    
+    private var breadcrumbId = Date().millisecondsSince1970
```
vlussenburg, line 5: Probably call this `lastBreadcrumbId` and auto increment when using it

Sources/Public/BacktraceBreadcrumbsLogManager.swift
vlussenburg: See other comment on auto increment - should we auto increment in the beginning instead?

Sources/Public/BacktraceBreadcrumbsLogManager.swift
```diff
@@ -0,0 +1,52 @@
...
+    
+    private var backtraceQueueFileHelper: BacktraceQueueFileHelper?
+
+    public init(_ breadcrumbLogPath: String, maxQueueFileSizeBytes: Int) {
+        super.init()
+        self.backtraceQueueFileHelper = BacktraceQueueFileHelper(breadcrumbLogPath, maxQueueFileSizeBytes: maxQueueFileSizeBytes)
+    }
+    
+    public func addBreadcrumb(_ message: String,
```
vlussenburg, line 14: I suspect this is not thread safe. It seems like two threads could go in there at the same time. Not sure how this works in Swift, have to admit.

Sources/Public/BacktraceQueueFileHelper.swift
```diff
@@ -0,0 +1,98 @@
...
+                return false
+            }
+        }
+        return false
+    }
+    
+    func clear() -> Bool {
+        do {
+            let fileURL = URL(fileURLWithPath: breadcrumbLogDirectory)
+            try "".write(to: fileURL, atomically: false, encoding: .utf8)
```
vlussenburg, line 58: Is there a more efficient native way to truncate a file?

Sources/Public/BacktraceQueueFileHelper.swift
```diff
@@ -0,0 +1,98 @@
...
+        }
+        return nil
+    }
+    
+    func convertInfoIntoString(_ info: Any) -> String? {
+        if let theJSONData = try? JSONSerialization.data( withJSONObject: info, options: []) {
+            let text = String(data: theJSONData, encoding: .ascii)
+            return text
+        }
+        return nil
```
vlussenburg, line 96: Shouldn't this throw an exception? Nil doesn't seem like the right thing to return.

Sources/Public/BacktraceBreadcrumbsLogManager.swift
MinPD126: The releavance of lastBreadcrumbId is different, lastBreadcrumbId is used to bind it in report with key "breadcrumbs.lastId", So I think current change is good enough. As per other comment we made increment id irrespective of result of add breadcrumb.

Sources/Public/BacktraceQueueFileHelper.swift
MinPD126: Another way is by using class FileHandle but its available iOS 13 onwards, so with respect of backward compatiblity current code is good.

Sources/Public/BacktraceBreadcrumbsLogManager.swift
MinPD126: Add breadcrumb feature exposed through BacktraceClient.shared and BacktraceClient is singleton class. So its thread safe and work good.

vlussenburg: Singleton or not, multiple threads can access a singleton at the same time, eg, it can be invoked from a foreground and background thread at roughly the same time. Can we include a unit test that verifies this?

https://medium.com/swiftcairo/avoiding-race-conditions-in-swift-9ccef0ec0b26

Sources/Public/BacktraceQueueFileHelper.swift
vlussenburg: https://www.techotopia.com/index.php/Working_with_Files_in_Swift_on_iOS_8#Truncating_a_File

Sources/Public/BacktraceBreadcrumb.swift
MinPD126: Changes has been done and code committed

Sources/Public/BacktraceQueueFileHelper.swift
MinPD126: Changes has been done and code committed

Sources/Public/BacktraceQueueFileHelper.swift
```diff
@@ -0,0 +1,103 @@
...
+
+extension BacktraceQueueFileHelper {
+    
+    func getSavedContent() -> [[String:Any]]? {
+        do {
+            let fileURL = URL(fileURLWithPath: breadcrumbLogDirectory)
+            let content = try String(contentsOf: fileURL, encoding: .utf8)
+            return convertStringIntoInfo(content)
+        } catch {
+            BacktraceLogger.warning("\(error.localizedDescription) \nWhen try to read breadcrumbs log file")
```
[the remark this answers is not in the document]
MinPD126: This code is no longer requred, with adding Cassette.

Sources/Public/BacktraceBreadcrumbsLogManager.swift
MinPD126: Integrate Cassette for better performance and muti thread support.

Sources/Public/BacktraceQueueFileHelper.swift
```diff
@@ -0,0 +1,103 @@
...
+                return nil
+            }
+        }
+        return nil
+    }
+    
+    func convertInfoIntoString(_ info: Any) -> String? {
+        do {
+            let theJSONData = try JSONSerialization.data( withJSONObject: info, options: [])
+            let text = String(data: theJSONData, encoding: .ascii)
```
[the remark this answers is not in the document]
MinPD126: Changed to utf8

Sources/Features/Breadcrumb/BacktraceComponentListener.swift
```diff
@@ -0,0 +1,126 @@
...
+//MARK: Orientation Status Listener
+    
+    private func observeOrientationChange() {
+        NotificationCenter.default.addObserver(self, selector: #selector(notifyOrientationChange), name: UIDevice.orientationDidChangeNotification, object: nil)
+    }
+    
+    @objc private func notifyOrientationChange() {
+        switch UIDevice.current.orientation{
+            case .portrait, .portraitUpsideDown:
+                addBradcrumb("portrait")
```
vlussenburg, line 26: Typo: bradcrumb => breadcrumb

Sources/Features/Breadcrumb/BacktraceComponentListener.swift
```diff
@@ -0,0 +1,126 @@
...
+    private func addBradcrumb(_ orientation: String) {
+        let attributes = ["orientation": orientation]
+        let _ = BacktraceClient.shared?.addBreadcrumb("Configuration changed",
+                                                      attributes: attributes,
+                                                      type: .system,
+                                                      level: .info)
+    }
+    
+    
+//MARK: Memroy Status Listener
```
vlussenburg, line 43: Typo: Memroy => memory

Sources/Features/Breadcrumb/BacktraceComponentListener.swift
```diff
@@ -0,0 +1,126 @@
...
+    
+    private func getMemoryWarningText(_ memoryPressureEvent: DispatchSource.MemoryPressureEvent) -> String {
+        if memoryPressureEvent.rawValue == DispatchSource.MemoryPressureEvent.normal.rawValue {
+            return "Generic memory warning"
+        } else if memoryPressureEvent.rawValue == DispatchSource.MemoryPressureEvent.critical.rawValue {
+            return "TRIM MEMORY RUNNING CRITICAL"
+        } else if memoryPressureEvent.rawValue == DispatchSource.MemoryPressureEvent.warning.rawValue {
+            return "TRIM MEMORY RUNNING WARNING"
+        } else {
+            return "Generic memory warning"
```
vlussenburg, line 77: Should the first option and the default option be the same string? If so, should be omit the first if since the result is the same?

Sources/Features/Breadcrumb/BacktraceComponentListener.swift
```diff
@@ -0,0 +1,126 @@
...
+    
+    private func getMemoryWarningLevel(_ memoryPressureEvent: DispatchSource.MemoryPressureEvent) -> BacktraceBreadcrumbLevel {
+        if memoryPressureEvent.rawValue == DispatchSource.MemoryPressureEvent.normal.rawValue {
+            return .warning
+        } else if memoryPressureEvent.rawValue == DispatchSource.MemoryPressureEvent.critical.rawValue {
+            return .fatal
+        } else if memoryPressureEvent.rawValue == DispatchSource.MemoryPressureEvent.warning.rawValue {
+            return .error
+        } else {
+            return .warning
```
vlussenburg, line 89: Same question as for the message, the first if and default option being the same (`.warning`).

Sources/Features/Breadcrumb/BacktraceComponentListener.swift
MinPD126: With else part we have to return some generic message that's why the same message("Generic memory warning") picked up. Please do suggest any other message if you have.

Sources/Features/Breadcrumb/BacktraceComponentListener.swift
MinPD126: With else part we have to return BacktraceBreadcrumbLevel that's why choosen .warning. Please do suggest any other BacktraceBreadcrumbLevel if you have.

vlussenburg: `.info` or `.debug`?

Sources/Public/BacktraceClient.swift
```diff
@@ -150,13 +150,16 @@ extension BacktraceClient: BacktraceReporting {
...
-        guard let resource = try? reporter.generate(exception: exception,
+        guard var resource = try? reporter.generate(exception: exception,
                                                     attachmentPaths: attachmentPaths,
                                                     faultMessage: faultMessage) else {
             completion(BacktraceResult(.unknownError))
             return
         }
 
+//         Check breadcrumb and attach report
+        configuration.breadcrumbSetting.processReportBreadcrumbs(&resource)
```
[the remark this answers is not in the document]
vlussenburg: This is okay for beta.

Sources/Features/Breadcrumb/BacktraceBreadcrumbFileHelper.swift
```diff
@@ -0,0 +1,73 @@
+import Foundation
+import Cassette
+
+enum BacktraceBreadcrumbFileHelperError: Error {
+    case invalidFormat
+}
+
+@objc public class BacktraceBreadcrumbFileHelper: NSObject {
```
konraddysput, line 8: What do you think about renaming class to something like backtraceBreadcrumbFileManager? The class doesn't look like a helper class because we have a lot of business logic here that doesn't look like a typical helper.

Sources/Features/Breadcrumb/BacktraceBreadcrumbFileHelper.swift
```diff
@@ -0,0 +1,73 @@
...
+     
+    func addBreadcrumb(_ breadcrumb:[String:Any]) -> Bool {
+        do {
+            var text = try convertBreadcrumbIntoString(breadcrumb)
+            if queue?.isEmpty() == false {
+                text = "\n\n\(text)"
+            }
+            let textBytes = Data(text.utf8)
+            if textBytes.count > 4096 {
+                BacktraceLogger.error("We should not have a breadcrumb this big, this is a bug!")
```
konraddysput, line 44: I disagree! how about a situation when I as a user want to store really long breadcrumb manually? Is this still an internal bug? I don't think so. In this situation also ... does it make sense to just cut the breadcrumb content instead of removing the whole breadcrumb?

Sources/Features/Breadcrumb/BacktraceBreadcrumbsLogManager.swift
```diff
@@ -0,0 +1,50 @@
...
+                       attributes:[String:Any]? = nil,
+                       type: BacktraceBreadcrumbType,
+                       level: BacktraceBreadcrumbLevel) -> Bool {
+        let time = Date().millisecondsSince1970
+        var breadcrumb: [String: Any] = ["timestamp": time,
+                                         "id": breadcrumbId,
+                                         "level": level.info,
+                                         "type": type.info,
+                                         "message": message]
+        if let attributes = attributes, !attributes.keys.isEmpty {
```
konraddysput, line 24: do we need to have this for loop if we already have attributes in the format [String:Any]? Also - shouldn't we expect attributes in format `[String: String]` ?

Sources/Features/Breadcrumb/BacktraceBreadcrumbsLogManager.swift
```diff
@@ -0,0 +1,50 @@
...
+            for attribute in attributes {
+                attribInfo[attribute.key] = attribute.value
+            }
+            breadcrumb["attributes"] = attribInfo
+        }
+        breadcrumbId += 1
+        if let result = backtraceBreadcrumbFileHelper?.addBreadcrumb(breadcrumb), result == true {
+            return result
+        }
+        return false
```
konraddysput, line 35: hmm isn't it a condition
```
if (true) {
 return true;
} 
return false;
```

Shouldn't we just return the addBreadcrumb method result instead?

Sources/Features/Breadcrumb/BacktraceBreadcrumbsLogManager.swift
```diff
@@ -0,0 +1,50 @@
...
+    }
+    
+    func clear() -> Bool {
+        let result = backtraceBreadcrumbFileHelper?.clear()
+        if result == true {
+            breadcrumbId = 0;
+        }
+        return result ?? false
```
konraddysput, line 43: small nit:
wouldn't it be better to return `result == true`?

Sources/Features/Breadcrumb/BacktraceComponentListener.swift
```diff
@@ -0,0 +1,126 @@
...
+            case .portrait, .portraitUpsideDown:
+                addBreadcrumb("portrait")
+            case .landscapeLeft, .landscapeRight:
+                addBreadcrumb("landscape")
+            default:
+                print("unknown")
+        }
+    }
+    
+    private func addBreadcrumb(_ orientation: String) {
```
konraddysput, line 34: the name of the method gives you a generic information about the purpose of the method. I disagree - the addBreadcrumb method adds breadcrumb with the orientation setting

Sources/Public/BacktraceBreadcrumb.swift
```diff
@@ -0,0 +1,163 @@
...
+#endif
+    private func breadcrumbLogPath() throws -> String {
+        var fileURL = try FileManager.default.url(for: .documentDirectory, in: .userDomainMask, appropriateFor: nil, create: true)
+        fileURL.appendPathComponent(BacktraceBreadcrumb.breadcrumbLogFileName)
+        return fileURL.path
+    }
+    
+    public func enableBreadCrumbs(_ breadCrumbTypes: [BacktraceBreadcrumbType] = BacktraceBreadcrumbType.all , maxLogSize: Int = defaultMaxLogSize) {
+        do {
+#if os(iOS)
```
konraddysput, line 105: on Mac/tvOS/other OS than iOS - does this method do anything?  If no - should we give a user a way to initialize breadcrumbs at all on os different than iOS?

Sources/Public/BacktraceBreadcrumb.swift
```diff
@@ -0,0 +1,163 @@
...
+            if let _ = breadCrumbTypes.first(where: { $0.rawValue == BacktraceBreadcrumbType.system.rawValue }) {
+                backtraceComponentListener = BacktraceComponentListener()
+            }
+#endif
+        } catch {
+            BacktraceLogger.warning("\(error.localizedDescription) \nWhen enable breadcrumbs")
+        }
+    }
+    
+    public func disableBreadCrumbs() {
```
konraddysput, line 118: disableBreadcrumbs

konraddysput (commented):
Hey,

I'm not sure if I missed the code responsible for setting up breadcrumbs after session restart or we don't have it here. What is the problem? If we attach breadcrumbs to report and we received a crash then:
* if the app restarts, do we still have the correct path to the breadcrumb file in the crash 
* if we need to retry sending the crash - because of the internet connection problem on the app startup, does a new breadcrumb file override the old one so we will miss breadcrumbs?

Sources/Public/BacktraceBreadcrumb.swift
vlussenburg: Currently Mac and tvOS is out of scope - so on these platforms it's supposed to do nothing for this beta version.

Sources/Features/Breadcrumb/BacktraceBreadcrumbFileHelper.swift
```diff
@@ -0,0 +1,73 @@
...
+import Cassette
+
+enum BacktraceBreadcrumbFileHelperError: Error {
+    case invalidFormat
+}
+
+@objc public class BacktraceBreadcrumbFileHelper: NSObject {
+    
+    private var minimumQueueFileSizeBytes = 4096;
+    private var maxQueueFileSizeBytes = 0
```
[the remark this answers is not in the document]
vlussenburg: I added some comments, is it more clear?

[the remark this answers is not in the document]
konraddysput: It looks much better - thank you!

Sources/Features/Breadcrumb/BacktraceBreadcrumbFileHelper.swift
vlussenburg: Good point. In this case it's the mirror image of the same "helper" in android. So I'll keep it the same for now, but record a ticket to update both. Fair?

Sources/Features/Breadcrumb/BacktraceBreadcrumbFileHelper.swift
vlussenburg: Fair point, however, this exact logic is present in Android as well. I will add the improvement to a ticket to improve both.

Sources/Features/Breadcrumb/BacktraceBreadcrumbsLogManager.swift
vlussenburg: done

PR #79, opened 2022-07-14
Breadcrumbs further development

vlussenburg:
Feature - 
1. Introduce log level and category

2. Enable breadcrumb with settings

3. Allow to add only those breadcrumb whose level is either equal or higher from configured settings’s level.

4. Add Mac OS support for breadcrumb

5. Add notification observer.

- iOS - 
> Orientation - Observer triggered every time device orientation’s change
> Memory - Observer triggered every time when memory threshold reaches to warning and critical level.
> Battery - Observer triggered when battery level change   

- MacOS -
> Memory - Observer triggered every time when memory threshold reaches to warning and critical level.
> Battery - Its triggered when enable observer.   

Test report of add breadcrumb event 
**iOS** 
Device 
iPhone 12 iOS 14.8 - Capture orientation
iPhone 12 iOS 14 8

iPhone 8 iOS 13.2 - Capture orientation
iPhone 8 iOS 13 2

Simulator
iPhone 13 Pro Max iOS 15.5 - Capture orientation and memory status
iPhone 13 Pro Max iOS 15 2

iPhone SE (2nd Generation) iOS 15.5 - Capture orientation and memory status
iPhone SE (2nd Gen) iOS 15 2

**MacOS**
MacOS Monterey 12.2.1 - Capture battery level and memory status
MacOS Monterey 12 2 1

Screen Shot 2022-08-23 at 4 30 44 PM


Examples/Example-iOS/AppDelegate.swift
```diff
@@ -55,7 +53,7 @@ final class AppDelegate: UIResponder, UIApplicationDelegate {
             }
         }
         let attributes = ["My Attribute":"My Attribute Value"]
-        BacktraceClient.shared?.addBreadcrumb("My Native Breadcrumb",
+        _ = BacktraceClient.shared?.addBreadcrumb("My Native Breadcrumb",
```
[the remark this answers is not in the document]
MinPD126: To avoid warning "Result of call to 'addBreadcrumb(_:attributes:type:level:)' is unused" we add this code. Significance of _ is to remove unused warning.

Examples/Example-iOS-ObjC/AppDelegate.m
```diff
@@ -27,7 +27,7 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
                                                    reportsPerMin: 3
                                                    allowsAttachingDebugger: TRUE
                                                    detectOOM: FALSE];
-    [configuration enableBreadCrumbs:@[@(BacktraceBreadcrumbTypeManual), @(BacktraceBreadcrumbTypeLog)]];
+    //[configuration enableBreadcrumbs:@[@(BacktraceBreadcrumbTypeManual), @(BacktraceBreadcrumbTypeLog)]];
```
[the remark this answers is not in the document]
MinPD126: Resolved, line cumommented.

Examples/Example-iOS-ObjC/AppDelegate.m
```diff
@@ -27,7 +27,7 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
                                                    reportsPerMin: 3
                                                    allowsAttachingDebugger: TRUE
                                                    detectOOM: FALSE];
-    [configuration enableBreadCrumbs:@[@(BacktraceBreadcrumbTypeManual), @(BacktraceBreadcrumbTypeLog)]];
+    [configuration enableBreadcrumbs:@[@(BacktraceBreadcrumbTypeManual), @(BacktraceBreadcrumbTypeLog)]];
```
vlussenburg, line 30: This code doesn't work anymore, need to be adjusted to new API

Sources/Features/Attributes/AttributesProvider.swift
```diff
@@ -46,7 +49,7 @@ extension AttributesProvider: SignalContext {
     }
 
     var attachmentPaths: [String] {
-        return attachments.map(\.path)
+        return attachments.map(\.path) + attributesSources.map(\.attachments).reduce([], +).map(\.path)
```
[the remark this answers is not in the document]
vlussenburg: Can you elaborate more? I think you may be confusing the attribute provider with the attribute storage class

Examples/Example-iOS-ObjC/AppDelegate.m
MinPD126: Code updated

MinPD126 (approved):

Examples/Example-macOS-ObjC/AppDelegate.m
```diff
@@ -41,6 +43,11 @@ - (void)applicationDidFinishLaunching:(NSNotification *)aNotification {
     } @finally {
         
     }
+    NSDictionary *attributes = @{@"My Attribute":@"My Attribute Value"};
+    [BacktraceClient.shared addBreadcrumb:@"My Native Breadcrumb"
```
konraddysput, line 47: does it make sense to add that this is a native breadcrumb? All breadcrumbs are from the native layer here.

Examples/Example-iOS/AppDelegate.swift
konraddysput, line 56: does it make sense to assign a method result to nothing?

Sources/Features/Breadcrumb/BacktraceBreadcrumbFileHelper.swift
```diff
@@ -5,38 +5,38 @@ enum BacktraceBreadcrumbFileHelperError: Error {
...
+    /** CASQueueFile is not thread safe, so all interactions with it should be done synchronously through this DispathQueue */
+    private let dispatchQueue = DispatchQueue(label: "io.backtrace.BacktraceBreadcrumbFileHelper@\(UUID().uuidString)")
+
+    public init(_ breadcrumbSettings: BacktraceBreadcrumbSettings) throws {
         do {
-            self.queue = try CASQueueFile.init(path: breadcrumbLogDirectory)
+            self.queue = try CASQueueFile.init(path: breadcrumbSettings.getBreadcrumbLogPath().path)
         } catch {
             BacktraceLogger.error("\(error.localizedDescription) \nWhen enabling breadcrumbs")
             throw error
```
konraddysput, line 28: we shouldn't rethrow an error here. Why? It might cause an infinity loop of crashes. How? when the directory is not available on the startup or something unexpected happened to the directory, then the code will rethrow an exception. What can we do instead? Wouldn't it be better IF we return boolean from the init method and depends on the result, continue initialization or disable breadcrumbs?

Sources/Features/Breadcrumb/BacktraceBreadcrumbFileHelper.swift
```diff
@@ -52,18 +52,21 @@ enum BacktraceBreadcrumbFileHelperError: Error {
...
 
         do {
-            // Keep removing until there's enough space to add the new breadcrumb
-            while (queueByteSize() + textBytes.count) > maxQueueFileSizeBytes {
-                try queue.pop(1, error: ())
-            }
+            try dispatchQueue.sync {
+                // Keep removing until there's enough space to add the new breadcrumb (leaving 512 bytes room)
+                while (queueByteSize() + textBytes.count) > (maxQueueFileSizeBytes - 512) {
+                    try queue.pop(1, error: ())
```
konraddysput, line 65: Correct me if I'm wrong but in swift/objective-c passing a reference to error object will automatically set an error if an error occurred. With that being said - wouldn't it be better if we instead using try/catch block use an exception reference? 
In the documentation:
https://github.com/linkedin/cassette
I also see that the pop/add method returns boolean - which tells me if we could add a text to the queue. Wouldn't it be better to use boolean instead of try/catch block? Right now try/catch block is useless

Sources/Features/Breadcrumb/BacktraceBreadcrumbFileHelper.swift
```diff
@@ -74,7 +77,9 @@ enum BacktraceBreadcrumbFileHelperError: Error {
 
     func clear() -> Bool {
         do {
-            try queue.clearAndReturnError()
+            try dispatchQueue.sync {
+                try queue.clearAndReturnError()
```
konraddysput, line 81: clearAndReturnError returns boolean - maybe it's better to use boolean instead?

Sources/Features/Breadcrumb/BacktraceBreadcrumbSettings.swift
```diff
@@ -0,0 +1,54 @@
...
+        self.maxIndividualBreadcrumbSizeBytes = maxIndividualBreadcrumbSizeBytes
+        self.maxQueueFileSizeBytes = maxQueueFileSizeBytes
+        self.breadcrumbLogFileName = breadcrumbLogFileName
+        self.breadcrumbTypes = breadcrumbTypes
+        self.breadcrumbLevel = breadcrumbLevel
+        super.init()
+    }
+
+    @objc public func getBreadcrumbLogPath() throws -> URL {
+        var fileURL = try FileManager.default.url(for: .documentDirectory, in: .userDomainMask, appropriateFor: nil, create: true)
```
konraddysput, line 50: wouldn't it be better to initialize the path in the constructor instead of here? Why I think it's better to have it in constructor ? We don't respect possible URL exception anywhere.

Sources/Features/Breadcrumb/BacktraceBreadcrumbs.swift
```diff
@@ -0,0 +1,128 @@
...
+            self.breadcrumbTypes = breadcrumbSettings.breadcrumbTypes
+            self.breadcrumbsLogManager = try BacktraceBreadcrumbsLogManager(breadcrumbSettings: breadcrumbSettings)
+            if breadcrumbSettings.breadcrumbTypes.contains(where: { $0.rawValue == BacktraceBreadcrumbType.system.rawValue }) {
+                backtraceNotificationObserver = BacktraceNotificationObserver(breadcrumbs: self)
+                backtraceNotificationObserver?.enableNotificationObserver()
+            }
+
+            try BreadcrumbsInfo.breadcrumbFile = breadcrumbSettings.getBreadcrumbLogPath()
+
+            isBreadcrumbsEnabled = true
```
konraddysput, line 81: why do we enable breadcrumbs if the getBreadcrumbLogPath fails?

Sources/Features/Breadcrumb/BacktraceBreadcrumbs.swift
```diff
@@ -0,0 +1,128 @@
...
+    private(set) var isBreadcrumbsEnabled: Bool = false
+
+    public func enableBreadcrumbs(_ breadcrumbSettings: BacktraceBreadcrumbSettings = BacktraceBreadcrumbSettings()) {
+        do {
+            self.breadcrumbLevel = breadcrumbSettings.breadcrumbLevel
+            self.breadcrumbTypes = breadcrumbSettings.breadcrumbTypes
+            self.breadcrumbsLogManager = try BacktraceBreadcrumbsLogManager(breadcrumbSettings: breadcrumbSettings)
+            if breadcrumbSettings.breadcrumbTypes.contains(where: { $0.rawValue == BacktraceBreadcrumbType.system.rawValue }) {
+                backtraceNotificationObserver = BacktraceNotificationObserver(breadcrumbs: self)
+                backtraceNotificationObserver?.enableNotificationObserver()
```
konraddysput, line 76: is there a chance that the backtraceNotificationObserver will be nullable?

Sources/Features/Breadcrumb/BacktraceBreadcrumbs.swift
```diff
@@ -0,0 +1,128 @@
...
+
+@objc public class BacktraceBreadcrumbs: NSObject {
+
+    private var breadcrumbsLogManager: BacktraceBreadcrumbsLogManager?
+    private var backtraceNotificationObserver: BacktraceNotificationObserver?
+    private var breadcrumbLevel: BacktraceBreadcrumbLevel?
+    private var breadcrumbTypes: [BacktraceBreadcrumbType]?
+    private(set) var isBreadcrumbsEnabled: Bool = false
+
+    public func enableBreadcrumbs(_ breadcrumbSettings: BacktraceBreadcrumbSettings = BacktraceBreadcrumbSettings()) {
```
konraddysput, line 69: shouldn't we return information if we successfully enable breadcrumbs integration? It would be great if our users can somehow prepare themself for situation when the interface is disabled.

Sources/Features/Breadcrumb/BacktraceBreadcrumbs.swift
```diff
@@ -0,0 +1,128 @@
...
+
+        // Remove currentBreadcrumbsId, which prevents it from being added
+        BreadcrumbsInfo.currentBreadcrumbsId = nil
+    }
+
+    func addBreadcrumb(_ message: String,
+                       attributes: [String: String]? = nil,
+                       type: BacktraceBreadcrumbType = BacktraceBreadcrumbType.manual,
+                       level: BacktraceBreadcrumbLevel = BacktraceBreadcrumbLevel.info) -> Bool {
+        if let breadcrumbsLogManager = breadcrumbsLogManager, allowBreadcrumbsToAdd(level) {
```
konraddysput, line 107: what if the breadcrumbsLogManager is disabled because the enable method threw an internal error and we disabled breadcrumbs? We got a null reference exception here.

Sources/Features/Breadcrumb/BacktraceBreadcrumbsLogManager.swift
```diff
@@ -2,39 +2,46 @@ import Foundation
...
 
-    private var breadcrumbId = Date().millisecondsSince1970
+    private var breadcrumbId: Int
     private let backtraceBreadcrumbFileHelper: BacktraceBreadcrumbFileHelper
 
-    init(_ breadcrumbLogPath: String, maxQueueFileSizeBytes: Int) throws {
-        self.backtraceBreadcrumbFileHelper = try BacktraceBreadcrumbFileHelper(breadcrumbLogPath,
-                                                                               maxQueueFileSizeBytes: maxQueueFileSizeBytes)
+    init(breadcrumbSettings: BacktraceBreadcrumbSettings) throws {
+        self.backtraceBreadcrumbFileHelper = try BacktraceBreadcrumbFileHelper(breadcrumbSettings)
```
konraddysput, line 9: What if the backtraceBreadcrumbFileHelper threw an exception? I think we shouldn't continue setting rest information.

Sources/Features/Breadcrumb/BacktraceNotificationObserver.swift
```diff
@@ -0,0 +1,277 @@
+import Foundation
```
konraddysput, line 1: why we don't remove observers? I couldn't find the code responsible for doing it.

konraddysput (changes requested):
I think we need to adjust our API and think about possible situations when the code path might fail before our users reach out to us with information that "something doesn't work"

Sources/Features/Breadcrumb/BacktraceBreadcrumbFileHelper.swift
vlussenburg: This is part of the ObjC/Swift magic.

```
Objective-C methods that return BOOL and return an NSError by reference are exposed to Swift as methods that use the try/catch error-handling mechanism, and return Void.
```

https://stackoverflow.com/questions/46626571/how-to-convert-objective-c-bool-to-swifts-bool-with-nserror-as-one-of-the-par

Sources/Features/Breadcrumb/BacktraceBreadcrumbFileHelper.swift
vlussenburg: Agreed. I think it may be better to return nil here (you can do that in Swift).

Examples/Example-iOS/AppDelegate.swift
vlussenburg: This is how to ignore a return value in Swift:
https://developer.apple.com/forums/thread/71943

Examples/Example-macOS-ObjC/AppDelegate.m
vlussenburg: It's just the example app, but we can adjust it

Sources/Features/Breadcrumb/BacktraceBreadcrumbFileHelper.swift
vlussenburg: This is part of the ObjC/Swift magic. The boolean is not exposed.

```
Objective-C methods that return BOOL and return an NSError by reference are exposed to Swift as methods that use the try/catch error-handling mechanism, and return Void.
```

https://stackoverflow.com/questions/46626571/how-to-convert-objective-c-bool-to-swifts-bool-with-nserror-as-one-of-the-par

Sources/Features/Breadcrumb/BacktraceBreadcrumbs.swift
vlussenburg: On line 76, it's pretty much guaranteed to not be null

Sources/Features/Breadcrumb/BacktraceBreadcrumbs.swift
vlussenburg: Although that is a good point, it's the same design as BacktraceMetrics: https://github.com/backtrace-labs/backtrace-cocoa/blob/44e30fa1c9095ddd61090bae9b9aa659185c2b62/Sources/Features/Metrics/BacktraceMetrics.swift#L42

Since the class still can be used (addBreadcrumbs just returns false, but doesn't caused exceptions) I would argue this is okay.

Sources/Features/Breadcrumb/BacktraceNotificationObserver.swift
vlussenburg: If you read the documentation on observers, it mentions that from iOS9 it is unregistered automatically:

```
If your app targets iOS 9.0 and later or macOS 10.11 and later, and you used [addObserver(_:selector:name:object:)](https://developer.apple.com/documentation/foundation/notificationcenter/1415360-addobserver), you do not need to unregister the observer. If you forget or are unable to remove the observer, the system cleans up the next time it would have posted to it.
```
https://developer.apple.com/documentation/foundation/notificationcenter/1413994-removeobserver

Sources/Features/Breadcrumb/BacktraceBreadcrumbsLogManager.swift
vlussenburg: Try without question mark throws the exception, so it's the constructing class' problem

Sources/Features/Breadcrumb/BacktraceBreadcrumbs.swift
vlussenburg: We don't. Try without question mark throws the exception.

Examples/Example-macOS-ObjC/AppDelegate.m
vlussenburg: Done.

Sources/Features/Breadcrumb/BacktraceBreadcrumbFileHelper.swift
vlussenburg: Actually, I think it's less graceful that way. Swift DOES allow you to return nil, but then we get in the situation the nil needs to be handled in the invoking class (LogManager). I think the change should be to not catch the error. Then all the problems constructing the dependencies are capture in BacktraceBreadcrumbs and they are left turned off.

Does that work? I'll push that change now.

Sources/Features/Breadcrumb/BacktraceBreadcrumbs.swift
vlussenburg: Also the user can invoke isBreadcrumbsEnabled property. We can make a ticket to adjust this in both APIs in the future though

Sources/Features/Breadcrumb/BacktraceBreadcrumbFileHelper.swift
vlussenburg: Done

Sources/Features/Breadcrumb/BacktraceBreadcrumbs.swift
vlussenburg: It could be:
```
let backtraceNotificationObserver = BacktraceNotificationObserver(breadcrumbs: self) backtraceNotificationObserver.enableNotificationObserver()
self.backtraceNotificationObserver = backtraceNotificationObserver
```
I prefer the current code.

WDYT?

Sources/Features/Breadcrumb/BacktraceBreadcrumbs.swift
vlussenburg: Actually Swift is really cool and prevents that via optional unwrapping. If it's null it falls through to `return false`

```
Unwrap an optional type with optional binding
Optional binding is similar to using an if-else block. The only subtle difference is that if the optional value is not nil, the unwrapped value is assigned to a new constant and further operations are performed on the constant.

You can do this using the if-let statement:

var password:String? = "$tr0ngp@$$w0rd"

if let unwrappedpass = password {
	print("Password is \(unwrappedpass)") //Password is $tr0ngp@$$w0rd
}
```
https://www.freecodecamp.org/news/optional-types-in-swift/

Sources/Features/Breadcrumb/BacktraceNotificationObserver.swift
vlussenburg: If you check BacktraceBreadcrumbs it nils the BacktraceNotificationObserver. Garbage collection of iOS will take care of it. 
https://docs.swift.org/swift-book/LanguageGuide/Deinitialization.html

Regardless, I spotted the opportunity to change the collection to a constant.

konraddysput (approved):

Sources/Features/Breadcrumb/BacktraceBreadcrumbSettings.swift
vlussenburg: I agree.

konst-sauce (approved):
I think this is ok for the initial beta version, there is still plenty of work plus I'll need to profile it.
I've checked all 5 examples - all of them send Breadcrumbs correctly. 
Polished Keys class a bit, so final URLs are formatted (universe and token are set once and no need to repeat them again for the submissionURL).

PR #94, opened 2022-09-01
Feature/simulator

konst-sauce:
Initial commit for 10 most common errors issue. 

Note: ObjC examples are still to be polished - they are on the way.

I've created a separate module for simulations - BacktraceSimulator. Also added a mock invisible XIB with primitive UI to test UI-related errors (like calling UI element on non-main thread or non-optional IBOutlet is nil (f.e. developer forgot to drag a reference in Interface Builder)).

Architecture is primitive, though flexible. Simulator cases are represented by a struct with case function as pointers and case title (to be shown in UI in a table view). We can add new simulation cases easily just by adding them to an array of available simulations (see BacktraceSimulator init()) or commenting existing cases to remove them.

Also merged some code a bit to reduce duplicated code - added to the new Common-Swift folder (same follows up for ObjC examples - will be in Common-ObjC folder (renamed from Common folder)).

Code review is not requested due to work in progress state.


.gitignore
```diff
@@ -6,6 +6,8 @@
 build/
 DerivedData/
 
+Examples/Konstantin/
```
vlussenburg, line 9: Probably not needed :)

Examples/Common-ObjC/Keys.m
```diff
@@ -5,7 +5,7 @@ @implementation Keys
 + (NSString const *) universeName {
     // Set your universe name here - https://support.backtrace.io/hc/en-us/articles/360040516451-What-is-a-submission-url-
     // e.g.: yourteam
-    return @"yourteam";
+    return @"konst-ryab";
```
vlussenburg, line 8: Would edit this out :)

Examples/Common-ObjC/Keys.m
```diff
@@ -16,7 +16,7 @@ + (NSString const *) backtraceUrl {
 
 + (NSString const *) backtraceToken {
     // Set your submission token - https://support.backtrace.io/hc/en-us/articles/360040105172
-    return @"token";
+    return @"04aecb6b3da05e8d83f2a27f9b5f41352ac792428c5e5f6a94ff58d43ee14e46";
```
vlussenburg, line 19: Same here.

Examples/Common-Swift/BacktraceSimulator.swift
```diff
@@ -0,0 +1,170 @@
+//
+//  BacktraceSimulator.swift
+//  Backtrace-iOS
+//
+
+import UIKit
+import Backtrace
+
+
```
vlussenburg, line 9: I think double empty lines create warnings in Xcode

Examples/Common-Swift/BacktraceSimulator.swift
```diff
@@ -0,0 +1,170 @@
...
+    func caseTitle(atIndex index: Int) -> String?
+    func executeCase(atIndex index: Int) throws
+}
+
+
+class BacktraceSimulator: BacktraceSimulatorProtocol {
+    
+    // MARK: Error cases
+    
+    // 1. Non-optional is nil and we call some method
```
vlussenburg, line 21: Would add comments on top of the relevant function

Examples/Common-Swift/BacktraceSimulator.swift
```diff
@@ -0,0 +1,170 @@
...
+
+    
+    // MARK: Variables
+    
+    fileprivate typealias CaseFunction = () -> ()
+    
+    private var availableCases: [ErrorCase] = []
+    private var mockView: MockView?
+    
+//    ProcessInfo
```
vlussenburg, line 49: Inconsistent `//` placement

Examples/Example-macOS-ObjC/ViewController.m
```diff
@@ -12,6 +13,8 @@ @implementation ViewController
 - (void)viewDidLoad {
     [super viewDidLoad];
 
+//    BacktraceSimulator * simulator = [[BacktraceSimulator alloc] init];
```
vlussenburg, line 16: Remove commented out code

.gitignore
konst-sauce: I've created my own folder for some staff which shouldn't be pushed (like token and universe name) - but forgot to remove it in Keys class - so pushed anyway :-)

vlussenburg: Fair. Could you call it something that is not your name, like `LocalConfiguration` or something? Then we could all do it that way :) Not a biggie if not.

Examples/Example-iOS-PnP/ViewControllerPnP.swift
```diff
@@ -134,43 +135,49 @@ extension ViewControllerPnP : AVPlayerViewControllerDelegate {
...
     }
 
 //    @available(iOS 8.0, *)
     func playerViewControllerDidStopPictureInPicture(_ playerViewController: AVPlayerViewController) {
-        
+        print("playerViewControllerDidStopPictureInPicture")
+//        present(playerVC!, animated: true) {}
     }
 
 //    @available(iOS 8.0, *)
```
vlussenburg, line 163: Why is this commented?

Examples/Example-iOS-PnP/ViewControllerPnP.swift
```diff
@@ -41,8 +44,9 @@ class ViewControllerPnP : UIViewController {
                 .commit()
         }
         
+        avPlayer?.actionAtItemEnd = .pause
         playerVC?.player = avPlayer
-
+//        playerVC?.canStartPictureInPictureAutomaticallyFromInline = true
```
vlussenburg, line 49: why is this commented?

PR #100, opened 2022-10-03
Feature/crashloop

konst-sauce:
Please review the current version of a Crash Loop Detector.
This request doesn't include a separate file fix to work with Unity+Cocoa environment.


Tests/BacktraceCrashLoopDetectorTests.swift
```diff
@@ -0,0 +1,52 @@
...
+
+final class BacktraceCrashLoopDetectorTests: QuickSpec {
+
+    override func spec() {
+        describe("Crash Loop Detector") {
+
+            let crashLoopDetector = BacktraceCrashLoopDetector()
+            let eventsCount = BacktraceCrashLoopDetector.consecutiveCrashesThreshold
+            let timeIntervalStep = 200
+            var isCrashLoop = false
```
vlussenburg, line 15: Why define it here? I would just keep it local to the tests. There's no reason to persist across test.

Sources/Public/BacktraceCrashLoopDetector.swift
```diff
@@ -0,0 +1,138 @@
...
+import Foundation
+
+@objc public class BacktraceCrashLoopDetector: NSObject {
+        
+    internal struct StartUpEvent: Codable {
+        var timestamp: Double
+        var isSuccessful: Bool
+    }
+    
+    @objc private static let plistKey = "CrashLoopDB"
```
vlussenburg, line 15: I don't like this name, since the file contains more than CrashLoops. It contains successful runs as well.

Sources/Public/BacktraceCrashLoopDetector.swift
```diff
@@ -0,0 +1,138 @@
...
+        let bundleIDPLCR = "com.plausiblelabs.crashreporter.data"
+        let crashReportDir = cacheDir.appendingPathComponent(bundleIDPLCR).appendingPathComponent(appIDPath)
+
+        let reportName = "live_report.plcrash"
+        let reportFullPath = crashReportDir.appendingPathComponent(reportName).absoluteString.replacingOccurrences(of: "file://", with: "")
+        print("reportFullPath: \(reportFullPath)")
+        return reportFullPath
+    }
+    
+    @objc private func hasCrashReport() -> Bool {
```
vlussenburg, line 75: This doesn't address OutOfMemory crashes, since this won't leave a `live_report.plcrash`. See BacktraceOomWatcher for which files that watcher writes.

Sources/Public/BacktraceCrashLoopDetector.swift
konst-sauce: Changed to "CrashLoopDetectorData"

Sources/Public/BacktraceCrashLoopDetector.swift
konst-sauce: Added to Trello

Sources/Public/BacktraceCrashLoopDetector.swift
```diff
@@ -0,0 +1,141 @@
...
+        // Thats why it formats path to PLCrashReporter's report file itself,
+        // for not to use PLCrashReporter's APIs at all
+        
+        let bundleIDBT = Bundle.main.bundleIdentifier ?? ""
+        let appIDPath = bundleIDBT.replacingOccurrences(of: "/", with: "_")
+
+        let paths = NSSearchPathForDirectoriesInDomains(.cachesDirectory, .userDomainMask, true)
+        let cacheDir = URL(fileURLWithPath: paths.isEmpty ? "" : paths[0])
+
+        let bundleIDPLCR = "com.plausiblelabs.crashreporter.data"
```
konraddysput, line 71: it isn't different for old plcrashreporter version and the one that Microsoft owns?

Sources/Public/BacktraceCrashLoopDetector.swift
```diff
@@ -0,0 +1,141 @@
...
+        BacktraceLogger.debug("reportFullPath: \(reportFullPath)")
+        return reportFullPath
+    }
+
+    @objc private func hasCrashReport() -> Bool {
+        let exists = FileManager.default.fileExists(atPath: reportFilePath())
+        return exists
+    }
+
+    @objc private func addCurrentEvent() {
```
konraddysput, line 88: what is current and why it is important to point out this is a current event? It is always a current event no matter what happen :P  Maybe renaming it to add or addStartupEvent will make more sense here?

Sources/Public/BacktraceCrashLoopDetector.swift
```diff
@@ -0,0 +1,141 @@
...
+        startupEvents.append(event)
+        
+        if startupEvents.count > BacktraceCrashLoopDetector.consecutiveCrashesThreshold {
+            startupEvents.remove(at: 0)
+        }
+
+        BacktraceLogger.debug("Event Added: \(startupEvents.count)")
+    }
+
+    @objc private func badEventsCount() -> Int {
```
konraddysput, line 103: badEventsCount doesn't describe also what it does - right now we're counting crashes. How about renaming it to startupCrashCounter() or something like that?

konraddysput: we can also pass here as a parameter crashCounter and ask this function to check if we are in the crash loop. By doing that we do not need to iterate over the whole array (it's still small anyway and it doesn't cost as anyway, but the flow might be cleaner?)

Sources/Public/BacktraceCrashLoopDetector.swift
```diff
@@ -0,0 +1,138 @@
...
+        saveEvents()
+
+        // true -> crash loop detected -> set safe mode
+        // false -> crash loop NOT detected -> set normal mode
+        let result = count >= BacktraceCrashLoopDetector.consecutiveCrashesThreshold
+        print("Finishing Crash Loop Detection: \(result)")
+        return result
+    }
+    
+    @objc func deleteCrashReport() {
```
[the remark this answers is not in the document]
konraddysput: where do we use it?

Sources/Public/BacktraceCrashLoopDetector.swift
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    @objc private func badEventsCount() -> Int {
+        
+        var badEventsCount = 0
+        for event in startupEvents.reversed() {
+            if event.isSuccessful {
+                break
+            }
+            badEventsCount += 1
```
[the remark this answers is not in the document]
konraddysput: we shouldn't test private functions imo

konraddysput (approved):

Sources/Public/BacktraceCrashLoopDetector.swift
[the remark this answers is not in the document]
konst-sauce: marked as obsolete for now

Examples/Example-iOS/AppDelegate.swift
```diff
@@ -18,6 +18,19 @@ final class AppDelegate: UIResponder, UIApplicationDelegate {
     
     func application(_ application: UIApplication,
                      didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
+        
+        // Enable crash loop detector, it crashes count threshold is not specified - default will be used
```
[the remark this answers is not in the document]
konst-sauce: updated

Sources/Public/BacktraceCrashLoopDetector.swift
konst-sauce: yes, but it is outdated now - because we rely on counter file

Sources/Public/BacktraceCrashLoopDetector.swift
konst-sauce: hm - basically I agree - name is a bit confusing - polished out

Sources/Public/BacktraceCrashLoopDetector.swift
konst-sauce: 1. Agree, changed it to 'consecutiveEventsCount', as crashes should be consecutive to detect crash loop.
2. The function itself calculates only how many consecutive events happened during last N (threshold) launches. Detection of a crash loop is done inside detectCrashLoop() (it loads events, adds current event, saves events, calculates consecutive, compares them with threshold and returns true or false. 
3. I will elaborate documentation and comments a bit, so it will be less confusing :P

Sources/Public/Internal/CrashLoopDetector/BacktraceCrashLoopCounter.swift
```diff
@@ -0,0 +1,79 @@
+//
+//  BacktraceCrashLoopCounter.swift
+//  Backtrace
+//
+
+import Foundation
+
+@objc internal class BacktraceCrashLoopCounter: BacktraceCrashLoop {
```
konraddysput, line 8: 1. Do we reset the counter somewhere? Let's pretend we have 4 random crashes in the last month. Do we keep 4 as a counter all the time and we wait until the 5th crash happens?
2. What we do in a situation when the session ends with a crash but it wasn't a startup crash? Its a really popular use case - In this situation we have something that we can call "crash loop" from the code perspective, however, the developer shouldn't worry about this situation.

Examples/Example-macOS-ObjC/Base.lproj/Main.storyboard
```diff
@@ -710,16 +711,16 @@
...
                                     <constraint firstAttribute="height" constant="50" id="wS4-BS-4BZ"/>
                                 </constraints>
-                                <buttonCell key="cell" type="push" title="Button" bezelStyle="rounded" alignment="center" borderStyle="border" imageScaling="proportionallyDown" inset="2" id="vpD-Xg-4cH">
+                                <buttonCell key="cell" type="push" title="Crash Me" bezelStyle="rounded" alignment="center" borderStyle="border" imageScaling="proportionallyDown" inset="2" id="vpD-Xg-4cH">
                                     <behavior key="behavior" pushIn="YES" lightByBackground="YES" lightByGray="YES"/>
                                     <font key="font" metaFont="system"/>
                                 </buttonCell>
                                 <connections>
-                                    <action selector="liveReportButtonAction:" target="XfG-lQ-9wD" id="r4X-yE-pnP"/>
+                                    <action selector="crashAction:" target="XfG-lQ-9wD" id="OfJ-Ce-bpL"/>
```
konraddysput, line 723: shouldn't liveReportButton stay here?

Examples/Example-iOS-ObjC/AppDelegate.m
```diff
@@ -10,6 +10,21 @@ @interface AppDelegate () <BacktraceClientDelegate>
 @implementation AppDelegate
 
 - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions {
+    
+    
+    /*  Enable crash loop detector.
```
konraddysput, line 15: Wouldn't be better if we describe the story what someone want to use it?

Examples/Example-iOS-ObjC/Info.plist
```diff
@@ -15,9 +15,9 @@
 	<key>CFBundlePackageType</key>
 	<string>APPL</string>
 	<key>CFBundleShortVersionString</key>
-	<string>$(MARKETING_VERSION)</string>
+	<string>1.0</string>
```
konraddysput, line 18: is this something that we should replace with 1.0 version? I think they're overridden by cicd?

Sources/Public/Internal/CrashLoopDetector/BacktraceCrashLoopDetector.swift
```diff
@@ -0,0 +1,180 @@
...
+        saveEvents()
+
+        consecutiveCrashesCount = consecutiveEventsCount()
+
+        /*
+            true -> crash loop detected -> set safe mode
+            false -> crash loop NOT detected -> set normal mode
+         */
+        let result = consecutiveCrashesCount >= BacktraceCrashLoopDetector.consecutiveCrashesThreshold
+        BacktraceCrashLoop.LogDebug("Finishing Crash Loop Detection: \(result)")
```
konraddysput, line 54: It is not important from the user perspective, maybe instead we can print something like "Is in the crash loop? \(result)" ?

Sources/Public/Internal/CrashLoopDetector/BacktraceCrashLoopDetector.swift
```diff
@@ -0,0 +1,180 @@
...
+
+    @available(*, deprecated, message: "Temporarily not needed")
+    @objc internal func reportFilePath() -> String {
+        
+        /*  Crash Loop Detector considers all other Backtrace modules as potentially dangerous.
+            Thats why it formats path to PLCrashReporter's report file itself,
+            for not to use PLCrashReporter's APIs at all
+         */
+        
+        let bundleIDBT = Bundle.main.bundleIdentifier ?? ""
```
konraddysput, line 149: small nit:

backtraceBundleId would be better I believe

Sources/Public/Internal/CrashLoopDetector/BacktraceCrashLoopDetector.swift
```diff
@@ -0,0 +1,180 @@
...
+        }
+        return string
+    }
+}
+
+// MARK: Deprecated methods
+extension BacktraceCrashLoopDetector {
+
+    @available(*, deprecated, message: "Temporarily not needed")
+    @objc internal func reportFilePath() -> String {
```
konraddysput, line 142: have you consider a situation when the crash report is from the previous session because we couldn't send it? To reproduce this problem disable internet on your phone and restart application twice. Would we increase the counter?

Sources/Public/Internal/CrashLoopDetector/BacktraceCrashLoopDetector.swift
```diff
@@ -0,0 +1,180 @@
...
+
+    @objc internal func clearStartupEvents() {
+        startupEvents.removeAll()
+        saveEvents()
+        BacktraceCrashLoop.LogDebug("Startup Events Cleared: \(startupEvents.count)")
+    }
+    
+    @objc internal func consecutiveEventsCount() -> Int {
+        
+        var count = 0
```
konraddysput, line 116: The addStartupEvent drops the first event if the number of events is greater than the counter. You might have a situation when you will have the following number of events with crashCount property
[4,5,1,2,3] for threshold 5

let's see what is the result of the algorithm for the example above
4, 5 is going to increase counter to 2
then 1,2,3 won't increase the counter.

so as you can see the counter should be equal to 3, but is equal to 2 instead. Not sure if I'm missing something but - there is a high chance to have this situation in production code. ALso the question related to the way how we increase counter and when we reset it is really relevant to this question

Examples/Example-iOS-ObjC/Info.plist
konst-sauce: Restored. Changed coz Simulator was shouting about bundle version.

Sources/Public/Internal/CrashLoopDetector/BacktraceCrashLoopDetector.swift
konst-sauce: modified

Examples/Example-iOS-ObjC/Info.plist
vlussenburg: Don’t do this :) read the original commit statement when I added this. Should not be a hardcoded version. Make sure you’re on a recent xcode version. Leave it on MARKETING_VERSION.

Sources/Public/Internal/CrashLoopDetector/BacktraceCrashLoopCounter.swift
```diff
@@ -0,0 +1,79 @@
...
+
+    static private var counter = 0
+
+    static internal func start() {
+        checkFileExists()
+        BacktraceCrashLoop.LogDebug("Cache Dir: \(cacheDir())")
+        BacktraceCrashLoop.LogDebug("Crash Loop Counter File Path: \(filePath())")
```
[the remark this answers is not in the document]
konst-sauce: agree, removed

Sources/Public/BacktraceClient.swift
```diff
@@ -85,10 +94,62 @@ import Foundation
...
+    }
+    
+    @objc public static func isInSafeMode() -> Bool {
+        return workingMode == .safe
+    }
+    
+    @objc public static func enableCrashLoopDetection(_ threshold: Int = 0) {
+        BacktraceCrashLoopCounter.start()
+
+        crashLoopDetector = BacktraceCrashLoopDetector()
```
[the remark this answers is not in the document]
konst-sauce: 1. Agree with passing to constructor - done.
2. We need it only at launch and using it only once, so there is no sense to keep it globally throughout the whole app and throughout the whole app's lifecycle - we don't know if final user counts every byte of memory, so using it as a local pointer just until necessary is a better way here, IMHO

Sources/Public/Internal/CrashLoopDetector/BacktraceCrashLoopCounter.swift
```diff
@@ -0,0 +1,79 @@
...
+        counter = Int(contents) ?? 0
+    }
+
+    static internal func crashesCount() -> Int {
+        return counter
+    }
+
+    static private func checkFileExists() {
+        let path = filePath()
+        /*
```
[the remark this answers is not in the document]
konst-sauce: Added it so ...fileExists check won't be removed accidentally or intentionally.

Sources/Public/Internal/CrashLoopDetector/BacktraceCrashLoopCounter.swift
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    static internal func reset() {
+        counter = 0
+        saveCounter()
+    }
+
+    static private func saveCounter() {
+        let contents = String(counter)
+        try? contents.write(toFile: filePath(), atomically: true, encoding: .utf8)
```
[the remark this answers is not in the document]
konst-sauce: You are right - but I have even better idea - remove writing counter to the file at all here - we can check PLCR's report file's creation data on next launch, but not sure how is it reliable - will do some tests.

Sources/Public/BacktraceClient.swift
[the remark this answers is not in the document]
konraddysput: `We need it only at launch and using it only once` - so it means if someone create one more class instance, we would have them two.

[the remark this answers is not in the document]
konst-sauce: wait -     private static var crashLoopDetector: BacktraceCrashLoopDetector? - it is static already and private, so it is like a singleton inside BacktraceClient

Sources/Public/Internal/CrashLoopDetector/BacktraceCrashLoopCounter.swift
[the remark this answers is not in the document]
konraddysput: how you know if the crash is from the previous session? you could have a situation when we couldn't send a crash and the crash is still available in the crash dir

Sources/Public/Internal/CrashLoopDetector/BacktraceCrashLoopDetector.swift
konst-sauce: That's why we have a counter, counter incremented every time crash happens and compared with previous value from user defaults upon next launch.

konst-sauce: counter increase only when crash happens, not when launch happens.
But - I'm reconsidering idea of counter - if we'll be checking crash file creation date at launch - counter wont be necessary.

Sources/Public/Internal/CrashLoopDetector/BacktraceCrashLoopCounter.swift
konst-sauce: 1. For now counter resets when we detect a crash loop. Yes we keep 4 as a counter all the time. 
2. I didn't come with idea how to differ startup crashes right upon launch and crash somewhere after some time after launch. But comparing timestamps between didFinishLaunch and PLCR's crash signal handler maybe (though again - we have limited time here and we'll need to persist result - what brings back necessity to make a file write operation)

Sources/Public/BacktraceClient.swift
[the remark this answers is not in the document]
konraddysput: you're creating a new class instance here. What will happen if you call it twice? I think you will override it. singleton would be when you check if the instance exists and if does exist then you're reusing it.

Sources/Public/Internal/CrashLoopDetector/BacktraceCrashLoopDetector.swift
konst-sauce: It fully duplicates PLCrashReporter's path creation, which uses main bundle, and crash report is stored under main bundle ID, not backtraces'

Sources/Public/Internal/CrashLoopDetector/BacktraceCrashLoopCounter.swift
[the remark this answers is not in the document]
konst-sauce: counter removed completely

Sources/Public/Internal/CrashLoopDetector/BacktraceCrashLoopDetector.swift
konst-sauce: In new logic - if file has same timestamp - it won't be considered as a crash loop continuation.

Sources/Public/Internal/CrashLoopDetector/BacktraceCrashLoopCounter.swift
konst-sauce: There Is no counter any more

PR #104, opened 2022-12-20
Proposed fix for sendWithError bug

konst-sauce:
This is my proposed quick fix for error.message not being changed bug. Please let me know your thoughts.

### DIFF
--- a/Sources/Public/BacktraceClient.swift
+++ b/Sources/Public/BacktraceClient.swift
@@ -161,6 +161,8 @@ extension BacktraceClient: BacktraceReporting {
             return
         }
 
+        resource.attributes["error.message"] = faultMessage
+        
         dispatcher.dispatch({ [weak self] in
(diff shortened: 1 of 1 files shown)


konraddysput:
@konst-sauce 

We do set it this way in the `BacktraceReporter` extension. This situation might happen in different flows - we should try to move it there. We might have a situation they use the flow that you don't support now. Please move setting error.message in the `BacktraceReporter` class.

In addition to that this diff doesn't solve this problem. We need to adjust AttributeProvider by including scoped attributes (that don't change over time like application name or application version) and dynamic attributes that we can set when the user wants to generate a report.

konraddysput:
@konst-sauce I also see that you try to merge this branch to release instead of to master. If you do it this way, you need to all merge all untested changes from the release branch to master, which means that there is more risk in this update. Can we please try to merge it to master instead?

PR #106, opened 2023-01-05
Release 1.7.5

konraddysput:
# Why

This diff changes a way how we read attributes during the report creation process. By changing mutable into immutable we can read error message every time when we create report attributes


konst-sauce:
Ok

konst-sauce (approved):

PR #107, opened 2023-03-29
Swift 5 support

melekr:
Migrate Backtrace-cocoa from swift 4.2 to swift 5
Update swift_version to ver 5.7
Update COCOAPODS to ver 1.12.0
Fix BacktraceLogger swift syntax

### DIFF
--- a/Podfile.lock
+++ b/Podfile.lock
@@ -1,5 +1,5 @@
 PODS:
-  - Backtrace (1.7.4-beta2):
+  - Backtrace (1.7.5):
     - Backtrace-PLCrashReporter (= 1.5.3)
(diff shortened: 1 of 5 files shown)


Podfile.lock
```diff
@@ -1,5 +1,5 @@
 PODS:
-  - Backtrace (1.7.4-beta2):
+  - Backtrace (1.7.5):
```
konraddysput, line 2: can we add `-beta` to version? We don't want anyone to treat is as official version. I also think we should bump the version to 1.8.0

konraddysput (changes requested):

konraddysput (approved):

PR #111, opened 2023-04-28
Native crumbs

melekr:


Sources/Features/Breadcrumb/BacktraceBreadcrumbFileHelper.swift
```diff
@@ -36,78 +32,75 @@ enum BacktraceBreadcrumbFileHelperError: Error {
...
+            for index in (0..<queue.count).reversed() {
+                guard let queueBreadcrumb = queue.allElements()[index] as? [String: Any] else {
+                    BacktraceLogger.warning("Error weh fetching breacrumbs from queue")
+                    return false
+                }
+                guard let breadcrumbSize = queueBreadcrumb["size"] as? Int else {
+                    BacktraceLogger.warning("Error when adding breadcrumbSize to array")
+                    return false
+                }
+                // Pop last element if size id greater than maxQueueFileSizeBytes
```
melekr, line 69: typo

Sources/Features/Breadcrumb/BacktraceBreadcrumbFileHelper.swift
```diff
@@ -36,78 +32,75 @@ enum BacktraceBreadcrumbFileHelperError: Error {
...
+                    }
+                    breadcrumbsArray.append(breadcrumbJsonData)
+                    size += breadcrumbSize
                 }
-
-                try queue.add(textBytes, error: ())
             }
+            // Write breadcrumbs to file
+            let breadcrumbString = "[\(breadcrumbsArray.joined(separator: ","))]"
+            try breadcrumbString.write(to: self.breadcrumbLogURL, atomically: true, encoding: .utf8)
```
melekr, line 83: this can be moved to another file alongside clearBreadcrumbLogFile method.

Sources/Features/Breadcrumb/BacktraceBreadcrumbFileHelper.swift
```diff
@@ -36,78 +32,75 @@ enum BacktraceBreadcrumbFileHelperError: Error {
...
-        do {
-            try dispatchQueue.sync {
-                try queue.clearAndReturnError()
-            }
-        } catch {
-            BacktraceLogger.warning("\(error.localizedDescription) \nWhen clearing breadcrumb file")
-            return false
+        dispatchQueue.sync {
+            queue.clear()
+            clearBreadcrumbLogFile(at:self.breadcrumbLogURL)
```
rick-bt, line 94: It's ok to remove the try/catch here?

Sources/Features/Breadcrumb/BacktraceBreadcrumbFileHelper.swift
```diff
@@ -36,78 +32,75 @@ enum BacktraceBreadcrumbFileHelperError: Error {
...
         do {
-            try dispatchQueue.sync {
-                // Keep removing until there's enough space to add the new breadcrumb (leaving 512 bytes room)
-                while (queueByteSize() + textBytes.count) > (maxQueueFileSizeBytes - 512) {
-                    try queue.pop(1, error: ())
+            // Serialize breadcrumb: [String: Any] into Data
+            let breadcrumbJsonData = try JSONSerialization.data(withJSONObject: breadcrumb)
+            // Serialize Data into a JSON string
+            guard let breadcrumbJsonString = String(data: breadcrumbJsonData, encoding: .utf8) else {
+                BacktraceLogger.warning("Error when adding breadcrumb to file")
```
melekr, line 42: warning messages can be improved

Sources/Features/Breadcrumb/BacktraceBreadcrumbFileHelper.swift
melekr: yes since clearBreadcrumbLogFile() has it now.

Sources/Features/Breadcrumb/BacktraceBreadcrumbFileHelper.swift
```diff
@@ -36,78 +32,75 @@ enum BacktraceBreadcrumbFileHelperError: Error {
...
         } catch {
-            BacktraceLogger.warning("\(error.localizedDescription) \nWhen adding breadcrumb to file")
+            BacktraceLogger.warning("Error when adding breadcrumb to file: \(error)")
             return false
         }
-
-        return true
     }
-
+
```
KrzaQ, line 90: ☝️

KrzaQ (approved):
I don't have much to say, it looks ok to me.

Sources/Features/Breadcrumb/BacktraceBreadcrumbFileHelper.swift
```diff
@@ -1,30 +1,26 @@
...
 
-    /** CASQueueFile is not thread safe, so all interactions with it should be done synchronously through this DispathQueue */
     private let dispatchQueue = DispatchQueue(label: "io.backtrace.BacktraceBreadcrumbFileHelper@\(UUID().uuidString)")
 
     public init(_ breadcrumbSettings: BacktraceBreadcrumbSettings) throws {
-        self.queue = try CASQueueFile.init(path: breadcrumbSettings.getBreadcrumbLogPath().path)
-
+        
+        self.breadcrumbLogURL = try breadcrumbSettings.getBreadcrumbLogPath()
+        self.queue = Queue<Any>()
```
konraddysput, line 22: Does it need to be any?

Sources/Features/Breadcrumb/BacktraceBreadcrumbFileHelper.swift
```diff
@@ -1,30 +1,26 @@
...
+    private let breadcrumbLogURL: URL
 
-    /** CASQueueFile is not thread safe, so all interactions with it should be done synchronously through this DispathQueue */
     private let dispatchQueue = DispatchQueue(label: "io.backtrace.BacktraceBreadcrumbFileHelper@\(UUID().uuidString)")
 
     public init(_ breadcrumbSettings: BacktraceBreadcrumbSettings) throws {
-        self.queue = try CASQueueFile.init(path: breadcrumbSettings.getBreadcrumbLogPath().path)
-
+        
+        self.breadcrumbLogURL = try breadcrumbSettings.getBreadcrumbLogPath()
```
konraddysput, line 21: why it can throw? Maybe it would be better to throw an error faster and here pass only data needed by BacktraceBreadcrumbFileHelper?

Sources/Features/Breadcrumb/BacktraceBreadcrumbFileHelper.swift
```diff
@@ -1,30 +1,26 @@
 import Foundation
-import Cassette
 
 enum BacktraceBreadcrumbFileHelperError: Error {
     case invalidFormat
 }
 
 @objc class BacktraceBreadcrumbFileHelper: NSObject {
```
konraddysput, line 7: Does this name sound good for you? It doesnt look like a helper

Sources/Features/Breadcrumb/BacktraceBreadcrumbFileHelper.swift
```diff
@@ -36,78 +32,75 @@ enum BacktraceBreadcrumbFileHelperError: Error {
...
+            }
+            // Store breadcrumb Json String and size in Dictionary [String : Any]
+            let queueBreadcrumb = ["breadcrumbJson": breadcrumbJsonString, "size": breadcrumbSize] as [String : Any]
+            // Queue breacrumb
+            queue.enqueue(queueBreadcrumb)
+            // Iterate over the queue from newest to oldest breadcrumb and build an array of encoded strings
+            var breadcrumbsArray = [String]()
+            var size = 0
+            for index in (0..<queue.count).reversed() {
+                guard let queueBreadcrumb = queue.allElements()[index] as? [String: Any] else {
```
konraddysput, line 61: does it make sense to move queue.allElements before the for loop?

It does becaue meanwhile something can change while you're still iterating

Sources/Features/Breadcrumb/BacktraceBreadcrumbFileHelper.swift
```diff
@@ -36,78 +32,75 @@ enum BacktraceBreadcrumbFileHelperError: Error {
...
+                    BacktraceLogger.warning("Error weh fetching breacrumbs from queue")
+                    return false
+                }
+                guard let breadcrumbSize = queueBreadcrumb["size"] as? Int else {
```
konraddysput, line 65: can we provide strong types and avoid casting/using any?

Sources/Features/Breadcrumb/BacktraceBreadcrumbFileHelper.swift
```diff
@@ -36,78 +32,75 @@ enum BacktraceBreadcrumbFileHelperError: Error {
...
+                    BacktraceLogger.warning("Error when adding breadcrumbSize to array")
+                    return false
+                }
+                // Pop last element if size id greater than maxQueueFileSizeBytes
+                if size + breadcrumbSize > maxQueueFileSizeBytes && !queue.isEmpty {
```
konraddysput, line 70: we should break here - why? there will be a chance we omit one breadcrumb but add a few older one and this algorithm will be fine with it. It would be better if we will have breadcrumbs in the right order and we won't miss any

Sources/Features/Breadcrumb/BacktraceBreadcrumbFileHelper.swift
konraddysput, line 21: what if the file doesn't exist? we don't create it anywhere. We always write to a file. How about creating a start method that will do startup logic?

Sources/Features/Breadcrumb/QueueFile.swift
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    func pop() -> T? {
+        guard !elements.isEmpty else {
+                return nil
+            }
+        return elements.popLast()
+    }
+    
+    public func allElements() -> [T] {
```
konraddysput, line 30: wouldn't it be better to create indexAt method instead of allElements?

Sources/Features/Breadcrumb/BacktraceBreadcrumbFileHelper.swift
```diff
@@ -36,78 +32,75 @@ enum BacktraceBreadcrumbFileHelperError: Error {
...
+                    queue.pop()
```
konraddysput, line 71: you want to remove first element - are you sure your breadcrumb order is valid after pop?

Sources/Features/Breadcrumb/QueueFile.swift
```diff
@@ -0,0 +1,45 @@
...
+        } else {
+            return elements.removeFirst()
+        }
+    }
+
+    func peek() -> T? {
+        return elements.first
+    }
+
+    func pop() -> T? {
```
konraddysput, line 23: are we sure in the queue we should have a pop method at all?

Sources/Features/Breadcrumb/BacktraceBreadcrumbFileHelper.swift
```diff
@@ -36,78 +32,75 @@ enum BacktraceBreadcrumbFileHelperError: Error {
 
         super.init()
     }
-
+    
     func addBreadcrumb(_ breadcrumb: [String: Any]) -> Bool {
```
konraddysput, line 36: normally in our Sdks we also limit breadcrumbs based on the number - mostly we care about the number of breadcrumbs rather than the size. I don't think any logic related to it - do we have it in options?

Sources/Features/Breadcrumb/BacktraceBreadcrumbFileHelper.swift
melekr: generic types are always a good idea, I can change it to [String : Any] or [String : Int] ?

Sources/Features/Breadcrumb/BacktraceBreadcrumbFileHelper.swift
melekr: we can change it, drop a class name and I'll rename it.

Sources/Features/Breadcrumb/BacktraceBreadcrumbFileHelper.swift
melekr: I can move that

Sources/Features/Breadcrumb/BacktraceBreadcrumbFileHelper.swift
melekr: I'm not sure, I'll see what I can do here.

Sources/Features/Breadcrumb/QueueFile.swift
melekr: I can do that

Sources/Features/Breadcrumb/BacktraceBreadcrumbFileHelper.swift
melekr: might need to add extra logic to make sure indexes are valid after pop

Sources/Features/Breadcrumb/QueueFile.swift
melekr: just a remove last will do I think, you have another suggestion?

Sources/Features/Breadcrumb/BacktraceBreadcrumbFileHelper.swift
melekr: I can add a limit, can you elaborate on this one?

Sources/Features/Breadcrumb/BacktraceBreadcrumbFileHelper.swift
melekr: ok , adding a break here.

Sources/Features/Breadcrumb/BacktraceBreadcrumbFileHelper.swift
melekr: the file will always exist;
`getBreadcrumbLogPath()` method will create a file if doesn't exist

Sources/Features/Breadcrumb/BacktraceBreadcrumbFile.swift
```diff
@@ -0,0 +1,119 @@
...
+            self.maxQueueFileSizeBytes = BacktraceBreadcrumbFile.minimumQueueFileSizeBytes
+        } else {
+            self.maxQueueFileSizeBytes = breadcrumbSettings.maxQueueFileSizeBytes
+        }
+
+        super.init()
+    }
+
+    func addBreadcrumb(_ breadcrumb: [String: Any]) -> Bool {
+        do {
```
konraddysput, line 33: What can go wrong here? we added try/catch into whole function, but I'm not sure what might throw an error here

Sources/Features/Breadcrumb/BacktraceBreadcrumbFile.swift
```diff
@@ -0,0 +1,119 @@
...
+            // Serialize breadcrumb: [String: Any] into Data
```
konraddysput, line 34: do we need to add a comment we're serializing an object? The line below the comment explains everything. I think it's better to leave comments when you have complex logic that is not obvious and requires documentation. Serializing JSON is too obvious to document

Sources/Features/Breadcrumb/BacktraceBreadcrumbFile.swift
```diff
@@ -0,0 +1,119 @@
...
+            let breadcrumbJsonData = try JSONSerialization.data(withJSONObject: breadcrumb)
+            // Serialize Data into a JSON string
```
konraddysput, line 36: does it make sense to add this comment?

Sources/Features/Breadcrumb/BacktraceBreadcrumbFile.swift
```diff
@@ -0,0 +1,119 @@
...
+                            return false
+                        }
+                        breadcrumbsArray.append(breadcrumbJsonData)
+                        size += breadcrumbSize
+                    }
+                }
+                // Write breadcrumbs to file
+                let breadcrumbString = "[\(breadcrumbsArray.joined(separator: ","))]"
+                writeBreadcrumbToLogFile(breadcrumb: breadcrumbString, at: self.breadcrumbLogURL)
+                return true
```
konraddysput, line 84: you don't need this here - you're returning at line 90

Sources/Features/Breadcrumb/BacktraceBreadcrumbFile.swift
```diff
@@ -0,0 +1,119 @@
...
+                    }
+                    guard let breadcrumbSize = queueBreadcrumb["size"] as? Int else {
+                        BacktraceLogger.warning("Error when adding breadcrumbSize to array")
+                        return false
+                    }
+                    // Pop last element if size is greater than maxQueueFileSizeBytes
+                    if size + breadcrumbSize > maxQueueFileSizeBytes && !queue.isEmpty {
+                        while (index != 0) {
+                            _ = queue.pop(at: index)
+                        }
```
konraddysput, line 71: can you add `break` here? There is no need to do one more for iteration

Sources/Features/Breadcrumb/BacktraceBreadcrumbFile.swift
```diff
@@ -0,0 +1,119 @@
...
+                    } else {
```
konraddysput, line 72: if you add break to the if statement above, you don't need to have else and one more identation

Sources/Features/Breadcrumb/BacktraceBreadcrumbFile.swift
```diff
@@ -0,0 +1,119 @@
...
+            _ = dispatchQueue.sync {
+                let queuedBreadcrumbs = queue.allElements()
+                var breadcrumbsArray = [String]()
+                var size = 0
+                for index in (0..<queue.count).reversed() {
+                    guard let queueBreadcrumb = queuedBreadcrumbs[index] as? [String: Any] else {
+                        BacktraceLogger.warning("Error when fetching breacrumbs from queue")
+                        return false
+                    }
+                    guard let breadcrumbSize = queueBreadcrumb["size"] as? Int else {
```
konraddysput, line 63: can we check if queueBreadcrumb exists? We can avoid casting everything to as?

Sources/Features/Breadcrumb/BacktraceBreadcrumbFile.swift
```diff
@@ -0,0 +1,119 @@
...
+        do {
+            try breadcrumb.write(to: breadcrumbLogURL, atomically: true, encoding: .utf8)
+        } catch {
+            BacktraceLogger.warning("Error writing breadcrumb to log file at: \(breadcrumbLogURL) - \(error.localizedDescription)")
+        }
+    }
+
+    func clearBreadcrumbLogFile(at breadcrumbLogURL: URL) {
+        do {
+            try "".write(to: breadcrumbLogURL, atomically: false, encoding: .utf8)
```
konraddysput, line 114: what if someone deletes a file when you're generating breadcrumbs?

konraddysput (approved):

Sources/Features/Breadcrumb/BacktraceBreadcrumbFileHelper.swift
melekr: there is no good way to do this since BacktraceBreadcrumbSettings forces us to (try/unwrap), if for any reason the file is nil the files operations methods will catch the error and warn BacktraceLogger.

Sources/Features/Breadcrumb/BacktraceBreadcrumbFile.swift
melekr: removed :)

Sources/Features/Breadcrumb/BacktraceBreadcrumbFile.swift
```diff
@@ -31,54 +31,39 @@ enum BacktraceBreadcrumbFileError: Error {
...
                 for index in (0..<queue.count).reversed() {
-                    guard let queueBreadcrumb = queuedBreadcrumbs[index] as? [String: Any] else {
-                        BacktraceLogger.warning("Error when fetching breacrumbs from queue")
-                        return false
-                    }
-                    guard let breadcrumbSize = queueBreadcrumb["size"] as? Int else {
-                        BacktraceLogger.warning("Error when adding breadcrumbSize to array")
-                        return false
-                    }
+                    let queueBreadcrumb = queuedBreadcrumbs[index]
```
konraddysput: We receive index out of bounds when we access breadcrumbs with an invalid index, right? or will it return null?

Sources/Features/Breadcrumb/QueueFile.swift
```diff
@@ -1,5 +1,15 @@
 import Foundation
 
+struct BreadcrumbRecord {
```
konraddysput, line 3: I think it should be moved to a separated file. 
You implemented queue to be generic. However, putting BreadcrumbRecord definition in the Queue implementation sounds weird.

Sources/Features/Breadcrumb/QueueFile.swift
```diff
@@ -1,5 +1,15 @@
 import Foundation
```
konraddysput, line 1: Is it QueueFile? I believe its a generic queue instead. Can we rename the file to Queue or something like that?

konraddysput (approved):

Sources/Features/Breadcrumb/QueueFile.swift
melekr: moving it

Sources/Features/Breadcrumb/QueueFile.swift
melekr: Queue it is!

Sources/Features/Breadcrumb/BacktraceBreadcrumbFile.swift
melekr: the way the code is laid out inside DispatchQueue it shouldn't produce an invalid index.
To be extra sure I'm going to add a method that returns an optional element at index and guard that.

melekr: it shouldn't but I'll guard it

Tests/BacktraceBreadcrumbTests.swift
```diff
@@ -163,10 +163,6 @@ final class BacktraceBreadcrumbTests: QuickSpec {
                     var writeIndex = 0
                     while writeIndex < 1000 {
                         let text = "this is Breadcrumb number \(writeIndex)"
-                        // submit a task to the queue for background execution
-                        DispatchQueue.global().async(group: group, execute: {
-                            expect { breadcrumbs.addBreadcrumb(text) }.to(beTrue())
```
vlussenburg, line 168: Without this background task, this test doesn't do what is intended to do.

We had lots of problems with multithreaded behavior corrupting the file, so beware!

Tests/BacktraceBreadcrumbTests.swift
```diff
@@ -192,21 +188,10 @@ final class BacktraceBreadcrumbTests: QuickSpec {
...
-                        let expectedNumberOfMatches = writeIndex - wrapIndex - 1
-                        expect(matches).to(beGreaterThanOrEqualTo(expectedNumberOfMatches),
-                                           description: "Not enough (\(matches) out of \(expectedNumberOfMatches)) " +
-                                           "breadcrumb matches found in breadcrumbs file: \n\(breadcrumbText)")
-
                         let attr = try FileManager.default.attributesOfItem(atPath: self.breadcrumbLogPath(false))
                         let fileSize = attr[FileAttributeKey.size] as? Int
                         let requestedSize = settings.maxQueueFileSizeBytes
                         expect { fileSize }.to(beLessThanOrEqualTo(requestedSize))
-                        expect { fileSize }.to(beGreaterThanOrEqualTo(requestedSize - 1000))
```
vlussenburg, line 209: This check is important. I would consider adjusting it, not deleting it.

Tests/BacktraceBreadcrumbTests.swift
```diff
@@ -192,21 +188,10 @@ final class BacktraceBreadcrumbTests: QuickSpec {
                             }
                         }
 
-                        // Why the - 1?
-                        // Because one line is liable to get mangled by the wrapping - half will
-                        // be at the bottom and half at the top of the circular file.
-                        // Currently, we accept we lose this Breadcrumb in the UI - it will still be in the file
-                        // for manual inspection.
-                        let expectedNumberOfMatches = writeIndex - wrapIndex - 1
-                        expect(matches).to(beGreaterThanOrEqualTo(expectedNumberOfMatches),
```
vlussenburg, line 201: This check is important. I would consider adjusting it, not deleting it.

Backtrace.podspec
```diff
@@ -31,8 +31,6 @@ Pod::Spec.new do |s|
   s.osx.public_header_files = ["Backtrace-macOS/**/*.h*"]
   s.tvos.public_header_files = ["Backtrace-tvOS/**/*.h*"]
 
-  s.ios.dependency "Cassette", '1.0.0-beta5'
```
vlussenburg, line 34: Check if there's any superfluous `#if os(iOS) || os(OSX)` left in prod/test code. I know they were a bunch of 'm

vlussenburg:
Hey hey! 

Cool that that pesky Cassette is finally being removed :). I forgot where, but there's performance test results stored in Github or Hack somewhere that document for which file sizes the performance is what. 

This PR seems to be doing somewhat similar to what https://github.com/backtrace-labs/backtrace-cocoa/pull/95 was meant to do (but is outdated now). You may want to compare and then close it.

Thanks :)
Vincent

Tests/BacktraceBreadcrumbTests.swift
vlussenburg: This test is specifically there to make sure the file rolls over, so the number of expected matches are really important. Otherwise the entire test doesn't do anything the other (smaller) tests already do.

Tests/BacktraceBreadcrumbTests.swift
melekr: will do

Backtrace.podspec
melekr: I'll take a look :)

PR #115, opened 2023-10-19
Added by default application.session and application.version attributes

konraddysput:
# Why

This diff always adds application.session and application.version attributes by default. Previously metrics system was required to make them available

ref: BT-745


rick-bt (approved):
If you deem appropriate, could you also add backtrace.agent and backtrace.version?

perf2711 (approved):

konraddysput:
> If you deem appropriate, could you also add backtrace.agent and backtrace.version?

I can do it in the next pull request

PR #119, opened 2024-01-19
Breadcrumb overflow

konraddysput:
# Why

When the number of breadcrumbs added by the app requires breadcrumb cleanup, the application is stuck in the while loop. This diff should allow dropping necessary breadcrumb without infinity loop 

ref BT-2112


Sources/Features/Attributes/DefaultAttributes.swift
```diff
@@ -101,7 +101,7 @@ struct Device: AttributesSource {
     var immutable: [String: Any?] {
         return [
             "device.machine": try? System.machine(),
-            "device.model": try? System.model(),
+            "device.model": try? System.machine(),
```
perf2711, line 104: is this intended?

Sources/Features/Breadcrumb/BacktraceBreadcrumbFile.swift
```diff
@@ -60,9 +53,10 @@ enum BacktraceBreadcrumbFileError: Error {
                 let breadcrumbSize = queueBreadcrumb.size
                 // Pop last element if size is greater than maxQueueFileSizeBytes
                 if size + breadcrumbSize > maxQueueFileSizeBytes && !queue.isEmpty {
-                    while (index != 0) {
-                        _ = queue.pop(at: index)
+                    for breadcrumbsToRemove in (0...index).reversed() {
+                       _ = queue.pop(at: breadcrumbsToRemove)
                     }
+
```
perf2711, line 59: is there no `shift` or `slice` function?

perf2711 (approved):

PR #126, opened 2024-06-24
Add extension to the uploaded attachment

konraddysput:
# Why

This diff adds the extension to file attachments. Thanks to this, our users can correctly see files in the web debugger view and don't need to rename them every time they download attachments

### DIFF
--- a/Sources/Features/Network/MultipartRequest.swift
+++ b/Sources/Features/Network/MultipartRequest.swift
@@ -65,9 +65,9 @@ extension MultipartRequest {
         body.appendString("\r\n")
 
         // attachments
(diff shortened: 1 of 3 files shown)


Sources/Features/Network/MultipartRequest.swift
```diff
@@ -65,9 +65,9 @@ extension MultipartRequest {
         body.appendString("\r\n")
 
         // attachments
-        for attachment in report.attachmentPaths.compactMap(Attachment.init(filePath:)) {
+        for attachment in Set(report.attachmentPaths).compactMap(Attachment.init(filePath:)) {
```
melekr, line 68: nothing wrong here, just want to point that we're reducing an Array to Set.

melekr (approved):
lgtm

PR #127, opened 2024-06-24
Rename attributes and add them as default

konraddysput:
# Why
This pull request always adds metrics that we thought are "metrics specific". We always want to include the session id, application version application build in the crash reports.


melekr (commented):
lgtm

melekr (approved):
lgtm

PR #128, opened 2024-06-25
Attribute standarization

konraddysput:
# Why

While ago we detected some attributes on iOS don't use our system attribute convention. We fixed them in react-native code: https://github.com/backtrace-labs/backtrace-javascript/pull/242/files and I just detected we introduced them everywhere based on attributes from this library.

This pull request standardizes the way how we call vm.* attributes


melekr (approved):
lgtm

PR #131, opened 2024-07-04
Bugfix/metrics dispatcher cicd

melekr:
This diff contains the following updates:
- Fixes metrics, dispatcher and watcher Tests
- Fixes Tests failures when running fastlane actions
- Fixes issue with runner timeout when waiting for emulator
- Raises ios & tvos DEPLOYMENT_TARGET to 12.0 and resolves build warnings
- Sets ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES to use inherited properties and resolves build warnings
- Removes LocationInfo attributes

ref: BT-3005


Sources/Features/Attributes/DefaultAttributes.swift
```diff
@@ -199,10 +199,19 @@ struct NetworkInfo: AttributesSource {
 struct LocationInfo: AttributesSource {
 
     var mutable: [String: Any?] {
-        return [
-            "location.enabled": CLLocationManager.locationServicesEnabled(),
-            "location.authorization.status": CLLocationManager.authorizationStatus().name
-        ]
+        var mutableDict = [String: Any?]()
```
konraddysput, line 202: I think it won't work. The location information won't be returned on line 214 but will be generated later.

Tests/BacktraceMetricsTest.swift
```diff
@@ -16,35 +16,31 @@ final class BacktraceMetricsTests: QuickSpec {
...
-                metrics.addUniqueEvent(name: uniqueEventName)
-                expect { metrics.count }.to(equal(2))
+            
+            var metrics: BacktraceMetrics?
+            
+            context("Events operation") {
+                
+                beforeEach {
+                    metrics = BacktraceMetrics(api: backtraceApi)
+                    metrics?.enable(settings: BacktraceMetricsSettings())
```
konraddysput, line 26: enable method should generate a summed + unique event. Are you sure we're testing the right thing? Maybe we should clear summed events if we want to test unique? I think its worth testing if the default summed +unique events are generated.

Tests/BacktraceMetricsTest.swift
```diff
@@ -16,35 +16,31 @@ final class BacktraceMetricsTests: QuickSpec {
...
+                    metrics?.clearSummedEvents()
+                }
+                
+                it("clears the summed event after enabling and sending") {
+                    // Allow default events to be "sent" out
+                    expect { metrics?.count }.toEventually(equal(1))
```
konraddysput, line 32: why 1? I think its becuase the unique event is still there. Maybe we can do better and get all events and filter them by checking the event type? Maybe it is just enough if we move 1 to variable and call it expectedNumberOfEvents

Tests/BacktraceWatcherTests.swift
```diff
@@ -43,19 +43,15 @@ final class BacktraceWatcherTests: QuickSpec {
...
-                    watcher.enable()
                     watcher.resetTimer()
+                    watcher.enable()
 
-                    waitUntil(timeout: .seconds(dbSettings.retryInterval + 1)) { (done) in
-                        watcher.configureTimer(with: DispatchWorkItem(block: {
-                            done()
-                        }))
-                    }
+                    expect(watcher.timer).toNot(beNil())
```
konraddysput, line 54: Are we testing the same thing?

Backtrace.xcodeproj/project.pbxproj
```diff
@@ -2558,15 +2557,14 @@
...
-				TVOS_DEPLOYMENT_TARGET = 11.0;
+				TVOS_DEPLOYMENT_TARGET = 12.0;
 			};
 			name = Debug;
 		};
 		28F95BC622525DCC003936E0 /* Release */ = {
 			isa = XCBuildConfiguration;
 			baseConfigurationReference = A1D05FC2301097A7A62A1A64 /* Pods-Backtrace-tvOSTests.release.xcconfig */;
 			buildSettings = {
-				ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES = YES;
```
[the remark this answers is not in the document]
melekr: yes, it's still the same value just inherited.
We had a warning about this.

Sources/Features/Attributes/DefaultAttributes.swift
melekr: seemed to work when I tested, I'll double check.

Tests/BacktraceMetricsTest.swift
melekr: all I did was cleanup by moving metrics init and enable to beforeEach.
Tests premises remain unchanged.
I added clearSummedEvents to test but it's not needed at all. I'll revert tests to the way they were.

Tests/BacktraceWatcherTests.swift
melekr: the scope of the test is to make sure timer fires given retry behaviour is enabled.
truth is waitUntil is the worst way to test this and does not play well with hangs.
This approach aims to unblock us for the moment, I'm going to be reworking our worker logic with better/newer syntax.

Tests/BacktraceMetricsTest.swift
konraddysput: I think we're not testing the most important functionality, which is checking if the unique/summed event will be generated on the app startup. Maybe we can move cleanup method to each test and allow the test to decide what is the behavior under the test?

Tests/BacktraceMetricsTest.swift
konraddysput: I think we should verify it the number of summed events is 1. metrics.count will return number of events in the queue (unique + summed)

Tests/BacktraceMetricsTest.swift
```diff
@@ -16,35 +16,31 @@ final class BacktraceMetricsTests: QuickSpec {
...
+                }
+                
+                it("can add and store summed event") {
+                    metrics?.addSummedEvent(name: summedEventName)
+                    expect { metrics?.count }.to(equal(2))
```
konraddysput, line 37: I think the test is about to verify if the number of tests added to the queue will increase. I think that's the better way to verify that. We shouldn't think about hard coded numbers but rather about the behavior
```suggestion
                    let count =  metrics?.count
                    metrics?.addSummedEvent(name: summedEventName)
                    expect { metrics?.count }.to(equal(count + 1))
```

Also maybe it would be better to verify if the summedEventName is in the metrics list?

Tests/DispatcherTests.swift
```diff
@@ -8,17 +8,16 @@ final class DispatcherTests: QuickSpec {
...
-                expect(finished).toEventually(beTrue(), timeout: .seconds(5), pollInterval: .milliseconds(100))
-                expect(closureCalled).toEventually(beTrue(), timeout: .seconds(5), pollInterval: .milliseconds(100))
+            let dispatcher = Dispatcher()
+            var dispatched = false
+            context("Dispatcher operation") {
+                it("calls the completion closure") {
+                    dispatcher.dispatch({
+                        dispatched = true
+                    }, completion: {
+                        expect(dispatched).to(beTrue())
```
konraddysput, line 18: I think this test also verifies if this callback was fired. Are you 100% sure we don't need to test it?

Tests/BacktraceWatcherTests.swift
konraddysput: can we put this as a comment in the test?

konraddysput (approved):

PR #147, opened 2024-12-18
Improvement/attachments and requests

melekr:
This PR refactors the MultipartRequest and Attachments implementation to transition from in-memory multipart body construction to disk-based streaming using a temporary file and improve filenameExtension detection.

### Why?
- The previous implementation (in-memory NSMutableData) held the entire multipart body in memory, risking high memory usage and even crashes for large reports or numerous chunky attachments (which is limited by available RAM), streaming to disk allows the sdk to process larger reports and attachments reliably.
- Content-Length is now calculated from the size of the temporary file, avoiding the need to hold the entire payload in memory
- Modern MIME Type Detection that leverage UTType APIs for iOS 14+/tvOS/macOS 11+ with fallbacks for older platforms

### Changes
- Replace NSMutableData operation with streaming Data to a temporary file
- Use httpBodyStream to stream file content directly into the request body
- Use FileHandle to manage file writing
- Calculate file size after writing to ensure accurate HTTP header
- defer file handling
- Migrate Attachements to modern UniformTypeIdentifiers API, keep legacy support

ref: BT-5311


Examples/Example-iOS-ObjC/ViewController.m
```diff
@@ -38,5 +38,4 @@ - (IBAction) crashAction: (id) sender {
     array[1];
 }
 
-
```
konraddysput, line 41: small nit: 
it's better to not include changes that don't belong to the pull request or adds additional work for reviewers

Package.swift
```diff
@@ -6,7 +6,7 @@ let package = Package(
     name: "Backtrace",
     platforms: [
         .iOS(.v12),
-        .macOS(.v10_13),
+        .macOS(.v11),
```
konraddysput, line 9: why do we want to bump the macos target?

Sources/Public/BacktraceCrashReporter.swift
```diff
@@ -50,8 +50,13 @@ extension BacktraceCrashReporter: CrashReporting {
     func generateLiveReport(exception: NSException? = nil,
                             attributes: Attributes,
                             attachmentPaths: [String] = []) throws -> BacktraceReport {
+        let reportData: Data
```
konraddysput, line 53: why - generateLiveReport allows to pass nullable exception - why it matters?

Sources/Features/Network/MultipartRequest.swift
```diff
@@ -48,41 +48,59 @@ extension MultipartRequest {
         var multipartRequest = urlRequest
         let boundary = generateBoundaryString()
         multipartRequest.setValue("multipart/form-data; boundary=\(boundary)", forHTTPHeaderField: "Content-Type")
-
-        let boundaryPrefix = "--\(boundary)\r\n"
-        let body = NSMutableData()
+        
+        // temporary file to stream data
+        let tempURL = URL(fileURLWithPath: NSTemporaryDirectory()).appendingPathComponent(UUID().uuidString)
```
konraddysput, line 53: what happen to the file if the app is restarted while file is being created?

Sources/Features/Network/MultipartRequest.swift
```diff
@@ -48,41 +48,59 @@ extension MultipartRequest {
...
+        defer { fileHandle.closeFile() }
+        
         // attributes
         for attribute in report.attributes {
-            body.appendString(boundaryPrefix)
-            body.appendString("Content-Disposition: form-data; name=\"\(attribute.key)\"\r\n\r\n")
-            body.appendString("\(attribute.value)\r\n")
+            writeToFile(fileHandle, "--\(boundary)\r\n")
+            writeToFile(fileHandle, "Content-Disposition: form-data; name=\"\(attribute.key)\"\r\n\r\n")
+            writeToFile(fileHandle, "\(attribute.value)\r\n")
```
konraddysput, line 63: shouldn't we do one write operation instead of multiple? I know opening writing and closing cost resources and time

konraddysput (changes requested):
The pull request description doesn't tell me why exactly we need it. Can you please share more thoughts on this?

Examples/Example-iOS-ObjC/ViewController.m
melekr: oops

Package.swift
melekr: macos 11 is way past its end of life, this was a perfect opportunity to let it go.
Next bump should be : 
```
.iOS(.v13),
.macOS(.v12),
.tvOS(.v13)
```

Sources/Features/Network/MultipartRequest.swift
melekr: file is created in Temp directory managed by OS, if the app is restarted there is no guarantee the file is persisted but : 
- OOM : MultipartRequest will be retried/reconstructed from the already generated and persisted live report.
- BacktraceReport/Resource :  MultipartRequest will be retried/reconstructed from repository
- Watcher: MultipartRequest will be retried/reconstructed from repository

Sources/Features/Network/MultipartRequest.swift
melekr: My intention was multiple small write operations with batched writes to reduce system overhead, i'll look into it.

Sources/Features/Network/MultipartRequest.swift
melekr: unlike NSMutableData it's even possible to recover multipart request file! (not sure if necessary but can be a good excuse to consider for `// TODO: T16698 - Add retry logic`)

Sources/Features/Network/MultipartRequest.swift
konraddysput: hmm i think it would be faster if we create a string with attributes in memory and then do one write operation. There is no chance we will oom the app via string with attributes so we should be good to go

Sources/Features/Network/MultipartRequest.swift
```diff
@@ -48,41 +48,59 @@ extension MultipartRequest {
...
+        writeToFile(fileHandle, "Content-Type: application/octet-stream\r\n\r\n")
+        fileHandle.write(report.reportData)
+        writeToFile(fileHandle, "\r\n")
+        
         // attachments
         for attachment in Set(report.attachmentPaths).compactMap(Attachment.init(filePath:)) {
-            body.appendString(boundaryPrefix)
-            body.appendString("Content-Disposition: form-data; name=\"\(attachment.filename)\"; filename=\"\(attachment.filename)\"\r\n")
-            body.appendString("Content-Type: \(attachment.mimeType)\r\n\r\n")
-            body.append(attachment.data)
```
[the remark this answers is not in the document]
melekr: we can use a [Stream](https://developer.apple.com/documentation/foundation/inputstream).
Benefits:
- no longer holds attachment or report data in memory before writing since data is written directly to file handle as it's being read
- Data is never loaded entirely into memory, great for multiple and/or large attachments
Caviats:
- Memory and buffer management
- Error handling (incomplete streams, stream cannot be opened
- Small buffer chunks

Sources/Public/BacktraceCrashReporter.swift
melekr: for testing, will remove.

melekr:
Holding data in memory for attributes or attachments changes the purpose of writing into files.
A hybrid approach would be buffering data in chunks, using a buffered stream.

Package.swift
melekr: reverted

Sources/Features/Error/BacktraceError.swift
```diff
@@ -62,8 +66,18 @@ extension NetworkError {
...
     var localizedDescription: String {
         switch self {
-        case .malformedUrl(let url): return "Provided URL cannot be parsed: \(url)."
-        case .unknownError: return "Unknown error occurred."
+        case .malformedUrl(let url):
+            return "Provided URL cannot be parsed: \(url)."
+        case .fileCreationFailed(let url):
+            return "File Error occurred: \(url)."
+        case .fileWriteFailed:
+            return "File Write Error occurred."
```
konraddysput, line 74: Suggestion: 

I think file is to generic and might be weird from the developer perspective that the file means a temporary file during HTTP submission. My recommendation is to make it more explicit by saying "Temporary submission file" instead.

Sources/Features/Network/MultipartRequest.swift
```diff
@@ -48,41 +48,85 @@ extension MultipartRequest {
...
-            body.appendString("\r\n")
+        
+        // temporary file to stream data
+        let tempURL = URL(fileURLWithPath: NSTemporaryDirectory()).appendingPathComponent(UUID().uuidString)
+        
+        do {
+            let fileCreated = FileManager.default.createFile(atPath: tempURL.path, contents: nil, attributes: nil)
+            if !fileCreated {
+                throw HttpError.fileCreationFailed(tempURL)
```
konraddysput, line 58: it won't be captured by the L:111 - which means we will return a generic error rather this this one?

Sources/Features/Network/MultipartRequest.swift
```diff
@@ -48,41 +48,85 @@ extension MultipartRequest {
...
+            var attributesString = ""
+            for attribute in report.attributes {
+                attributesString += "--\(boundary)\r\n"
+                attributesString += "Content-Disposition: form-data; name=\"\(attribute.key)\"\r\n\r\n"
+                attributesString += "\(attribute.value)\r\n"
+            }
+            try writeToFile(fileHandle, attributesString)
+            
+            // report
+            var reportString = "--\(boundary)\r\n"
```
konraddysput, line 80: It would be better if we start with report rather than with attributes - if the application stops during submission part, we can still submit a valid report.

Sources/Features/Network/MultipartRequest.swift
```diff
@@ -48,41 +48,85 @@ extension MultipartRequest {
...
+            // attachments
+            for attachmentPath in Set(report.attachmentPaths) {
+                guard let attachment = Attachment(filePath: attachmentPath) else {
+                    BacktraceLogger.error("Failed to create attachment for path: \(attachmentPath)")
+                    continue
+                }
+                
+                try writeToFile(fileHandle, "--\(boundary)\r\n")
+                try writeToFile(fileHandle, "Content-Disposition: form-data; name=\"\(attachment.filename)\"; filename=\"\(attachment.filename)\"\r\n")
+                try writeToFile(fileHandle, "Content-Type: \(attachment.mimeType)\r\n\r\n")
```
konraddysput, line 96: I think we should do it with a single write

Sources/Features/Network/MultipartRequest.swift
```diff
@@ -48,41 +48,85 @@ extension MultipartRequest {
...
+                fileHandle.write(attachment.data)
```
konraddysput, line 97: you're making an attachment copy in the temporary file. WE cannot stream a file directory from the file available somewhere?

Sources/Features/Error/BacktraceError.swift
melekr: yes, it looks out of place
also need to properly handle an 'attachmentError'.

Sources/Features/Network/MultipartRequest.swift
melekr: possible but requires creating an InputStream for each attachments.
If we decide to keep files, i'll make the necessary improvements.

melekr:
- file handling, cleaning and integrity checks to be added id we decide to keep it.
- Error granularity improvements to be added

Sources/Features/Error/BacktraceError.swift
```diff
@@ -62,8 +66,18 @@ extension NetworkError {
 extension HttpError {
     var localizedDescription: String {
         switch self {
-        case .malformedUrl(let url): return "Provided URL cannot be parsed: \(url)."
-        case .unknownError: return "Unknown error occurred."
+        case .malformedUrl(let url):
+            return "Provided URL cannot be parsed: \(url)."
```
perf2711, line 70: small nit: can we either get rid of trailing dots or put the URL in quotes?
this goes for every description that has some data displayed in the string

Sources/Features/Network/MultipartRequest.swift
```diff
@@ -48,41 +48,85 @@ extension MultipartRequest {
...
+            
+            // Final boundary
+            try writeToFile(fileHandle, "--\(boundary)--\r\n")
+            
+            // Set Content-Length
+            let fileSize = try FileManager.default.attributesOfItem(atPath: tempURL.path)[.size] as? Int ?? 0
+            multipartRequest.setValue("\(fileSize)", forHTTPHeaderField: "Content-Length")
+            
+            // Attach file stream to HTTP body
+            multipartRequest.httpBodyStream = InputStream(url: tempURL)
```
perf2711, line 109: Instead of creating the file and setting it as `httpBodyStream`, couldn't you write directly to the request?

I'm not proficient in Swift, but I found this:
https://developer.apple.com/documentation/foundation/url_loading_system/uploading_streams_of_data

You can use bound streams to achieve that, it seems.

melekr: we can, this is the the approach I was referencing:
> Holding data in memory for attributes or attachments changes the purpose of writing into files. A hybrid approach would be buffering data in chunks, using a buffered stream.

Sources/Features/Network/MultipartRequest.swift
```diff
@@ -48,91 +48,118 @@ extension MultipartRequest {
...
             reportString += "Content-Disposition: form-data; name=\"upload_file\"; filename=\"upload_file\"\r\n"
             reportString += "Content-Type: application/octet-stream\r\n\r\n"
-            try writeToFile(fileHandle, reportString)
-            fileHandle.write(report.reportData)
-            try writeToFile(fileHandle, "\r\n")
-            
+            try writeToStream(outputStream, reportString, writeLock: writeLock)
+            try writeLock.sync {
+                let data = report.reportData
+                let bytesWritten = data.withUnsafeBytes {
```
konraddysput, line 77: why do we need to execute this part of the code in the sync context? You won't submit two identical reports.

Sources/Features/Network/MultipartRequest.swift
```diff
@@ -48,91 +48,118 @@ extension MultipartRequest {
...
+                try writeToStream(outputStream, "\r\n", writeLock: writeLock)
             }
-            
+
             // Final boundary
-            try writeToFile(fileHandle, "--\(boundary)--\r\n")
+            try writeToStream(outputStream, "--\(boundary)--\r\n", writeLock: writeLock)
             
+            // Data from Output Stream
+            guard let data = outputStream.property(forKey: .dataWrittenToMemoryStreamKey) as? Data else {
```
konraddysput, line 118: do we read data from the output stream here?

PR #151, opened 2025-01-28
Core Data Concurrency update

melekr:
#### Why?
When in concurrency debugging mode `-com.apple.CoreData.ConcurrencyDebug 1`; any direct call into the context methods from an unexpected thread or queue immediately triggers a concurrency violation.
This requires that a private queue context `.privateQueueConcurrencyType` must be accessed exclusively through its `perform(_:)` or `performAndWait(_:)` methods, which schedule all reads/writes on its private serial queue.

#### Changes:
- Add NSManagedObjectContext extension to capture any thrown error and return the result
- Update public Repository methods to use backgroundContext.performAndWaitThrowing
- Remove direct off-queue fetch/save calls to fix concurrency rule violations
- Move fetch/delete logic into private _locked helpers
- Add doc comments
- Expand Core Data Test coverage

ref: BT-5464

### DIFF
--- a/Sources/Features/Extensions/NSManagedObjectContext+Extensions.swift
+++ b/Sources/Features/Extensions/NSManagedObjectContext+Extensions.swift
@@ -1,0 +1,44 @@
+import Foundation
+import CoreData
+
+enum PerformAndWaitError: Error {
+    case blockDidNotRun
+}
+
+extension NSManagedObjectContext {
+
+    /// Runs `block` inside `performAndWait`, captures any thrown error and returns the result.
+    ///
+    /// - Parameter block: A closure that either returns `T` or throws an error
+    /// - Returns: The value returned by `block`
+    /// - Throws:
+    ///     - Rethrows any error from `block`, or `PerformAndWaitError.blockDidNotRun` if the closure never produced a result
+    func performAndWaitThrowing<T>(_ block: () throws -> T) throws -> T {
+        var result: T!
+        var thrownError: Error?
+        performAndWait {
+            do {
+                result = try block()
+            } catch {
+                thrownError = error
+            }
+        }
+        if let thrownError = thrownError {
+            throw thrownError
+        }
+        return result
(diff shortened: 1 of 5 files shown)


Sources/Features/Extensions/NSManagedObjectContext+Extensions.swift
```diff
@@ -0,0 +1,44 @@
...
+                thrownError = error
+            }
+        }
+        if let thrownError = thrownError {
+            throw thrownError
+        }
+        return result
+    }
+    
+    // Swift 5 approach
```
melekr, line 32: swift 5 approach is cleaner

Sources/Features/Extensions/NSManagedObjectContext+Extensions.swift
```diff
@@ -0,0 +1,44 @@
...
+    func performAndWaitThrowingSwift5<T>(_ block: () throws -> T) throws -> T {
+        var result: Swift.Result<T, Error>?
+            performAndWait {
+                // Captures returned value or throws error
+                result = Swift.Result(catching: block)
+            }
+            guard let outcome = result else {
+                throw PerformAndWaitError.blockDidNotRun
```
melekr, line 40: edge case

Tests/BacktraceDatabaseTests.swift
```diff
@@ -51,6 +51,83 @@ final class BacktraceDatabaseTests: QuickSpec {
...
+                                let report = try crashReporter.generateLiveReport(attributes: [:])
+                                try repository.save(report)
+                            } catch {
+                                fail("Failed to save concurrently: \(error)")
+                            }
+                        }
+                    }
+                    
+                    // concurrent reads
+                    for _ in 1...5 {
```
melekr, line 76: higher concurrent reads/writes take a lot of time.

konraddysput (approved):

PR #154, opened 2025-03-28
Update CPU, Architecture and OSInfo system attributes 

melekr:
Adds OSInfo name, version and buildNumber
Adds CPU/Device architecture
Fixes Uname .machine and .model attributes
Adds isSimulator attribute

ref: BT-5714

### DIFF
--- a/Sources/Features/Attributes/DefaultAttributes.swift
+++ b/Sources/Features/Attributes/DefaultAttributes.swift
@@ -119,10 +119,10 @@ struct Device: AttributesSource {
         let sysname = OSInfo.name
         result["uname.sysname"] = sysname
 
-        let machine = System.machine
+        let machine = try? System.machine()
         result["uname.machine"] = machine
 
-        let model = System.model
+        let model = try? System.model()
         result["device.model"] = model
 
         result["device.isSimulator"] = isSimulator()
--- a/Sources/Features/Attributes/DefaultAttributes.swift
+++ b/Sources/Features/Attributes/DefaultAttributes.swift
@@ -104,8 +104,9 @@ struct Device: AttributesSource {
     var immutable: [String : Any?] {
         var result: [String : Any?] = [:]
 
-        result["device.arch"] = CPU.architecture()
-        result["cpu.arch"] = CPU.architecture()
+        let architecture = CPU.architecture()
+        result["device.arch"] = architecture
+        result["cpu.arch"] = architecture
         
         let sysname = OSInfo.name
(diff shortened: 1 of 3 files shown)


Sources/Features/Attributes/DefaultAttributes.swift
```diff
@@ -101,12 +101,29 @@ struct Device: AttributesSource {
...
-        return [
-            "device.machine": try? System.machine(),
-            "device.model": try? System.machine(),
-            "uname.sysname": getSysname()
-        ]
+    var immutable: [String : Any?] {
+        var result: [String : Any?] = [:]
+
+        result["device.arch"] = CPU.architecture()
+        result["cpu.arch"] = CPU.architecture()
```
konraddysput, line 108: 1. why both?
2. you don't have any caching here - it means this function will do a lot of stuff twice.

Sources/Features/Attributes/DefaultAttributes.swift
```diff
@@ -101,12 +101,29 @@ struct Device: AttributesSource {
...
+        
+        let sysname = OSInfo.name
+        result["uname.sysname"] = sysname
+
+        let machine = try? System.machine()
+        result["uname.machine"] = machine
+        result["device.machine"] = machine
```
konraddysput, line 115: why both

Sources/Features/Attributes/System.swift
```diff
@@ -125,6 +127,33 @@ struct SystemControl {
             return baseAddress.withMemoryRebound(to: T.self, capacity: 1, { $0.pointee })
         })
     }
+    
+    static func string(forKeys keys: [Int32]) throws -> String {
```
konraddysput, line 131: what's the point of the string functioN?

konraddysput (changes requested):

Sources/Features/Attributes/DefaultAttributes.swift
melekr: setting both to avoid "Unknown Architecture", device.arch & cpu.arch are the same thing but developers can filter using either one.

Sources/Features/Attributes/DefaultAttributes.swift
melekr: same

Sources/Features/Attributes/System.swift
konraddysput: if we can change the function name it would be great.

konraddysput (approved):

PR #157, opened 2025-04-23
Update Example App and add CustomDirectoryBacktraceClientTests

melekr:


Examples/Example-iOS/AppDelegate.swift
```diff
@@ -11,35 +12,44 @@ func throwingFunc() throws {
...
-        BacktraceClient.shared?.delegate = self
         
-        do {
-            try throwingFunc()
-        } catch {
-            BacktraceClient.shared?.send(attachmentPaths: []) { (result) in
-                print("AppDelegate:Result:\(result)")
-            }
+        // Customize PLCrashReporterConfig with custom basePath https://docs.saucelabs.com/error-reporting/platform-integrations/ios/configuration/#plcrashreporter
+        guard let plcrashReporterConfig = PLCrashReporterConfig(signalHandlerType: .BSD, symbolicationStrategy: .all, basePath: crashDirectory.path) else {
```
konraddysput, line 36: what if we move to Crashpad? The dedicated settings object for it should be in the database options. This provides an abstraction layer on top of the crash reporting solution. We should stick to the database options - I'm against introducing "plcrashreporter specific" settings.

konraddysput (changes requested):

konraddysput (approved):

PR #158, opened 2025-06-02
Thread aware live-report capture

melekr:
### Why

No public APIs changes, Handled-exception reports now show the actual faulting thread in Backtrace, eliminating false “Thread 0/1 (CRASHED)” stacks.

- adds generateLiveReport with thread to CrashReporting protocol 
- defers mach_port_deallocate to avoid Mach-port leaks 
- generates reports with faulting thread
- Adds thread forwarding and leak spec tests

ref: BT-5583

### DIFF
--- a/Sources/Features/Client/BacktraceReporter.swift
+++ b/Sources/Features/Client/BacktraceReporter.swift
@@ -118,6 +118,7 @@ extension BacktraceReporter {
                   faultMessage: String? = nil) throws -> BacktraceReport {
         attributesProvider.set(faultMessage: faultMessage)
         let resource = try reporter.generateLiveReport(exception: exception,
+                                                       thread: mach_thread_self(),
                                                        attributes: attributesProvider.allAttributes,
                                                        attachmentPaths: attachmentPaths + attributesProvider.attachmentPaths)
         
--- a/Sources/Features/Client/BacktraceReporter.swift
+++ b/Sources/Features/Client/BacktraceReporter.swift
@@ -118,7 +118,6 @@ extension BacktraceReporter {
                   faultMessage: String? = nil) throws -> BacktraceReport {
         attributesProvider.set(faultMessage: faultMessage)
         let resource = try reporter.generateLiveReport(exception: exception,
-                                                       thread: mach_thread_self(),
                                                        attributes: attributesProvider.allAttributes,
                                                        attachmentPaths: attachmentPaths + attributesProvider.attachmentPaths)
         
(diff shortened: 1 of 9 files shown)


Sources/Features/Client/BacktraceReporter.swift
```diff
@@ -118,6 +118,7 @@ extension BacktraceReporter {
                   faultMessage: String? = nil) throws -> BacktraceReport {
         attributesProvider.set(faultMessage: faultMessage)
         let resource = try reporter.generateLiveReport(exception: exception,
+                                                       thread: mach_thread_self(),
```
konraddysput, line 121: Should we always use the main thread id instead?

Sources/Public/BacktraceCrashReporter.swift
```diff
@@ -54,6 +54,25 @@ extension BacktraceCrashReporter: CrashReporting {
         let reportData = try reporter.generateLiveReport(with: exception)
```
konraddysput, line 54: I think it would be better if you will stick to one implementation for everything. The generateLiveReport (no thread argument) can call generateLiveReport with a thread argument by using current thread. You don't need to have a duplicated logic. If you start extending it at some point, you would need to double your code

Sources/Public/BacktraceCrashReporter.swift
```diff
@@ -54,6 +54,25 @@ extension BacktraceCrashReporter: CrashReporting {
...
+        
+        defer { mach_port_deallocate(mach_task_self_, thread) }
+        let reportData = try reporter.generateLiveReport(withThread: thread, exception: exception)
+        
+        return try BacktraceReport(report: reportData, attributes: attributes, attachmentPaths: attachmentPaths)
+    }
+    
+    // Returns the thread ID
+    // can be used for logging/testing
+    static func currentThreadIdentifier() -> thread_t {
```
konraddysput, line 71: shouldn't this be a helper? We already had a lot of code in Common. This class should be as clean as possible since this is a business logic of our SDK

Sources/Features/Client/BacktraceReporter.swift
melekr: No, we want to capture the specific currently executing thread, the main thread would be missing context and/or reporting wrong informations if the error occurs on a background thread for example.

melekr: + callers who want main thread can still pass it explicitly.

Sources/Public/BacktraceCrashReporter.swift
melekr: Common is for example app, removing this method since it's useless after consolidating generateLiveReport methods.

Sources/Public/BacktraceCrashReporter.swift
melekr: sure thing, motivation was offering both methods without introducing a breaking change.

konraddysput (approved):

